PR #16, opened 2026-08-05
fix(sheet): preserve merged-cell spans that extend past the used range

### LINKED ISSUE #8: XLSX merged-cell spans are clipped to the populated range

mathurshubham:
Fixes #8.

## What

A merge anchored on populated cells but reaching into empty rows or columns
was clipped to the used range, so a valid source merge (`F1:O3` with only
`F1` populated) silently collapsed to an ordinary 1×1 cell in
`to_document()`.

The collapse happened in **two layers**, so both change:

1. **Sheet frontend** (`src/formats/sheet/mod.rs`) intersected each merged
   region with calamine's used range before building the grid — the clip the
   issue reports.
2. **`GridBuilder::finish`** (`src/model/table.rs`) dropped trailing rows
   holding only covered positions ("all-empty" treated covered slots as
   empty) and then re-clamped the surviving spans — so even an unclipped
   span would have shrunk back.

## The fix

- The sheet frontend keeps a region's **full extent**. The two existing
  defenses are preserved unchanged: the clamped top-left still decides the
  origin, and regions that don't intersect the used range at all are still
  ignored (the `merge_outside_the_used_columns_is_ignored` test is
  untouched). Rows below the used range are appended so the overhang has
  positions to materialize into.
- `finish()`'s trailing trim now exempts **span-claimed rows**: covered
  positions are structure a producer declared, and dropping them silently
  shrinks a valid span. Rows of plain empty cells still trim (new test pins
  that).

**Runaway regions stay bounded** — a region whose full extent would blow the
expansion budget (a crafted `A1:XFD1048576` merge, or a real whole-column
merge) degrades to today's clipped form instead of materializing: the file
still converts, it just loses the overhang (new test proves it).

## Behavior

The reporter's exact openpyxl repro, before → after:

```diff
-1 1
-origin 1 1
+3 10
+origin 3 10
```

The rendered **Markdown is unchanged** — the renderer already trims the
trailing empty area visually, so `to_markdown` still emits
`| Merged heading |`. The fix is observable through `to_document()`, which
is the surface the issue is about. Consequently **no existing snapshot
changed**.

## Tests

- Sheet: full-extent span over a 2×2 used range (5×5 merge), the issue's
  1×1-anchor shape (3×10 span, 29 covered), the existing out-of-range-merge
  defense unchanged, and the whole-sheet crafted merge degrading to the
  clipped form instead of erroring
  (`oversized_merge_degrades_to_the_clipped_form`).
- Model: trailing span-claimed rows survive `finish()` (replaces the test
  that asserted the old clamping behavior); trailing empty-origin rows still
  trim.
- Integration: new `handmade-overhang.xlsx` fixture + snapshot, and a
  model-retention test asserting the 3×10 grid and spans via
  `to_document()`.
- `cargo fmt --check`, `clippy --workspace --all-targets --all-features
  -D warnings`, and `cargo test --locked` all pass; only the new snapshot
  was added, nothing else moved.

## Note

This branch is rebased on current main and composes with the new
`resolve_header_rows` detection (fe97550) — header resolution runs on the
finished grid, including the preserved overhang rows.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/anydoc/pr/16"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788516109&installation_model_id=11126&pr_number=16&repository=firecrawl%2Fanydoc&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fanydoc%2Fpull%2F16&signature=a74ed3cbf97303a44cff452c631f672aa3364e8c36836025e5273beb22bb013b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


mathurshubham:
Closing in favor of #13, which was opened first and takes the same two-layer approach (sheet-side extent + the finish() trailing-trim) — I hadn't seen it when I opened this. 

One deliberate difference worth carrying over from this PR if useful: an integration fixture (handmade-overhang.xlsx) plus a to_document() retention test asserting the 3×10 grid and spans — happy to contribute those to #13 as a follow-up if wanted. Verified #13 against the issue's openpyxl repro locally; it resolves it.

mathurshubham:
Reopening: #13, which this was closed in favor of, was closed when its author's account was deleted, so #8 is unaddressed again. 

Since then I've folded in the one behavior where #13 had the better call — an oversized region (a crafted whole-sheet merge, or a real whole-column merge) now degrades to today's clipped form instead of failing the conversion with ResourceLimit (c19fbb0). 

Branch is rebased on current main and the full suite passes.

24arnabmaity-dev (approved):

PR #18, opened 2026-08-05
fix(csv): promote a label-like first row to the table header

### LINKED ISSUE #15: CSV first row is rendered as a body row under an empty header row
CSV conversion is row-for-row complete, but the first row is rendered as a *body* row underneath an empty GFM header row, so column names end up in the table body.

### Repro

`header.csv`:

```csv
name,role,team
Alice,Engineer,Platform
Bob,Designer,Product
```

`anydoc header.csv`:

```
|  |  |  |
| --- | --- | --- |
| name | role | team |
| Alice | Engineer | Platform |
| Bob | Designer | Product |
```

Expected (or at least, useful):

```
| name | role | team |
| --- | --- | --- |
| Alice | Engineer | Platform |
| Bob | Designer | Product |
```

### Cause

`formats/csv.rs` builds the table with `Table::from_rows(rows, 0, TableKind::Data)` — zero header rows — after configuring the reader with `.has_headers(false)`. The renderer still has to emit a header row because GFM requires one, so it emits an empty one.

The reasoning is sound: CSV has no header semantics, and guessing wrong would silently delete a data row. But the current output has real costs — the empty header renders as a visually blank row, and anything consuming the Markdown (an LLM, a table parser) sees the column names as data.

### Suggestions

- **Heuristic promotion**: treat row 0 as a header when it's fully populated, all-text, and every cell is distinct, while at least one later row parses as numeric/date. This is what `csv` sniffing tools do, and it fails safe to the current behavior.
- **Or make it explicit**: a `has_headers` option on the CSV path, defaulting to today's behavior.

The heuristic seems worth it given headerless CSVs are rarer in practice than headed ones, but an explicit opt-in avoids ever being wrong. Either would be an improvement over the blank row.

mathurshubham:
Fixes #15

This takes the heuristic direction from the issue, with one correction to
its framing that makes a simpler rule safe.

## What

CSV built its table with zero header rows, and GFM requires a header, so the
renderer emitted a blank header row with the column names sitting in the
body as data:

```
|  |  |  |
| --- | --- | --- |
| name | role | team |
| Alice | Engineer | Platform |
```

Now the first record is promoted to the header when it looks like labels:

```
| name | role | team |
| --- | --- | --- |
| Alice | Engineer | Platform |
```

## Why the heuristic can be simpler than the issue proposed

Two observations changed the design:

1. **A false positive only restyles a row.** Promotion sets `header_rows = 1`
   — the row still renders, in the header position. Nothing is deleted, so
   the issue's "guessing wrong would silently delete a data row" cost doesn't
   exist. The real asymmetry is: false positive = cosmetic; false negative =
   blank header + column names read as data.
2. **The issue's own suggested signal ("at least one later row parses as
   numeric/date") fails the issue's own repro** — `Alice,Engineer,Platform`
   is all text. Header detection can't rely on body types.

So the rule asks only that the first record *look like labels*, with no body
evidence required: **at least two columns, as wide as every record, and each
field non-empty, non-numeric (decimal commas like `1,5` count as numeric),
and distinct** (case-insensitive). Anything else — numeric first rows,
ragged or repeated fields, empty fields, a single record, one column —
fails safe to today's headerless output.

## Tests

- Unit: the issue's all-text repro promotes; numeric first rows (dot and
  decimal-comma) don't; duplicate labels, empty fields, and a first row
  narrower than the body don't; single-record and single-column files don't;
  promotion keeps all rows (nothing dropped).
- New fixture `handmade-numeric.csv` + snapshot pinning the fail-safe
  (numeric data keeps the blank header).
- All four existing CSV snapshots update to exactly the issue's "Expected"
  shape — header names in the header row, body unchanged. No non-CSV
  snapshot moved.
- `cargo fmt --check`, `clippy --workspace --all-targets --all-features
  -D warnings`, `cargo test --locked` all green.

## On the alternative

The issue also floated an explicit `has_headers` option. That's compatible
with this change (an option could override the heuristic later), but the
public API today takes bytes in and gives Markdown out with no per-format
options, so the heuristic route avoids inventing new API surface for the
common case.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/anydoc/pr/18"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788517493&installation_model_id=11126&pr_number=18&repository=firecrawl%2Fanydoc&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fanydoc%2Fpull%2F18&signature=0d035406e99ee17c3c88ee01eb6098faba05377764460efc79874ea872eba828"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


mathurshubham:
Closing — superseded by fe97550, which lands a shared type-vote header detector covering CSV, spreadsheets, and ODF. Verified the issue's all-text repro promotes correctly on main.

PR #21, opened 2026-08-05
feat(wasm): name the failure cause in a kind property

marcelolx:
A conversion error gives only a message. To tell an encrypted document from an unreadable one, a caller must match the message text. That match breaks when the wording changes.

This pull request adds a `kind` property to the error. The property holds the `ConvertError` variant. The values are `unsupported`, `malformed`, `encrypted`, `resourceLimit`, `missingPart`, and `io`. The message does not change, and it stays for people to read.

The error also takes the name `ConvertError`, which makes it easy to identify in a stack trace. The TypeScript definitions declare both types.

The `JsError` type cannot hold a property. Because of that limit, `toMarkdownBytes` and `toDocument` now return `Result<_, JsValue>`, and the code builds the error with `js_sys::Error`. This adds the `js-sys` crate, which is already part of the wasm-bindgen tree.

`ConvertError` is `#[non_exhaustive]`, so upstream can add a variant later. The `error_kind` function has a default branch, which keeps the crate correct when that happens.

`wasm/test.mjs` passes all 8 tests. One new test asserts the `kind` value twice: once for an input the code does not recognize, and once for an input that is malformed but named.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/anydoc/pr/21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788536410&installation_model_id=11126&pr_number=21&repository=firecrawl%2Fanydoc&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fanydoc%2Fpull%2F21&signature=694dad89db6b59c29c9d36db66f6bc5770533e64384dd54e3d73df718cce2cb6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -127,6 +127,7 @@ name = "anydoc-wasm"
 version = "0.1.4"
 dependencies = [
  "anydoc",
+ "js-sys",
  "serde",
  "serde-wasm-bindgen",
  "serde_bytes",
--- a/wasm/Cargo.toml
+++ b/wasm/Cargo.toml
@@ -21,6 +21,7 @@ wasm-opt = ["-O", "--enable-bulk-memory", "--enable-nontrapping-float-to-int"]
 
 [dependencies]
 anydoc = { path = ".." }
+js-sys = "0.3"
 serde = { version = "1", features = ["derive"] }
 serde_bytes = "0.11"
 serde-wasm-bindgen = "0.6"
--- a/wasm/src/lib.rs
+++ b/wasm/src/lib.rs
@@ -97,7 +97,7 @@ pub fn format_from_path(path: &str) -> Option<Format> {
 /// detected from the content, which signature-less formats (CSV) have to name
 /// explicitly.
 #[wasm_bindgen(js_name = toMarkdownBytes)]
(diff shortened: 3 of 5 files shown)


marcelolx:
@tomsideguide this one would be very helpful to have; we have been running a fork of anydoc that adds this property, so it would be really nice to see it land in the main project here, so we don't have to use our fork 🙏 

And thanks for adding wasm support!

tomsideguide:
@marcelolx this has been implemented in #25, released under v0.1.6

marcelolx:
Thanks @tomsideguide

Issue #26, opened 2026-08-05
docx: a mid-paragraph page break is dropped, concatenating the runs it separated


xmasyx:
A `<w:br w:type="page"/>` that sits **between two runs of text inside the same paragraph** is dropped without any replacement separator, so the words on either side are concatenated into one. `Alfa` + page break + `Beta` converts to `AlfaBeta`.

Markdown has no pages, so dropping the break itself is the right call. The loss here is the word boundary, not the page: the output is not merely unpaginated, it contains a word that was never in the document.

## Reproduction

```js
// npm i docx @firecrawl/anydoc
import { Document, Packer, Paragraph, PageBreak, TextRun } from 'docx';
import { toMarkdown } from '@firecrawl/anydoc';
import { writeFileSync } from 'node:fs';

const mk = (children) => new Document({ sections: [{ children }] });

const cases = {
  'mid-paragraph':      [new Paragraph({ children: [new TextRun('Alfa'), new PageBreak(), new TextRun('Beta')] })],
  'end of paragraph':   [new Paragraph({ children: [new TextRun('Alfa'), new PageBreak()] }), new Paragraph('Beta')],
  'own paragraph':      [new Paragraph('Alfa'), new Paragraph({ children: [new PageBreak()] }), new Paragraph('Beta')],
  'soft break (w:br)':  [new Paragraph({ children: [new TextRun('Alfa'), new TextRun({ text: 'Beta', break: 1 })] })],
};

for (const [name, children] of Object.entries(cases)) {
  const file = `/tmp/${name.replace(/\W+/g, '-')}.docx`;
  writeFileSync(file, await Packer.toBuffer(mk(children)));
  console.log(name.padEnd(20), JSON.stringify((await toMarkdown(file)).trim()));
}
```

Observed on `@firecrawl/anydoc@0.1.6`:

| case | output | |
| --- | --- | --- |
| mid-paragraph | `"AlfaBeta"` | ← the defect |
| end of paragraph | `"Alfa\n\nBeta"` | correct |
| own paragraph | `"Alfa\n\nBeta"` | correct |
| soft break (`w:br`) | `"Alfa\\\nBeta"` | correct |

**Expected:** `"Alfa\\\nBeta"` for the first case, i.e. the same treatment a soft break already gets.

## Scope

[...]

PR #32, opened 2026-08-06
fix(presentations): separate slides with a thematic break

### LINKED ISSUE #31: Presentation slides are concatenated with no boundary, so untitled slides merge into the previous one
## Summary

The three presentation parsers concatenate every slide's blocks into one list with nothing between them, so a slide with no title placeholder is indistinguishable from a continuation of the previous slide.

A slide's title becomes a `Heading`, so decks that title every slide read correctly by accident. A slide with no title contributes no structural block at all — two such slides in a row produce two adjacent paragraphs, exactly as if they were two paragraphs of one slide, and bullet lists from different slides merge into one list.

## Reproduction

Tested with `firecrawl-anydoc==0.1.6`.

```python
from pptx import Presentation
from pptx.util import Inches
import anydoc

prs = Presentation()
s1 = prs.slides.add_slide(prs.slide_layouts[5])
s1.shapes.title.text = "Titled slide"
for text in ("Second slide, no title placeholder.", "Third slide, also untitled."):
    slide = prs.slides.add_slide(prs.slide_layouts[6])   # blank layout
    slide.shapes.add_textbox(Inches(1), Inches(1), Inches(8), Inches(2)).text_frame.text = text
prs.save("untitled.pptx")

print(anydoc.to_markdown("untitled.pptx"))
```

## Observed

```
## Titled slide

Second slide, no title placeholder.

Third slide, also untitled.
```

Nothing marks where slide 2 ends and slide 3 begins.

## Expected

A separator between slides:

```
## Titled slide

---

Second slide, no title placeholder.

---

Third slide, also untitled.
```

## Why this is not the pagination you already declined

#26 settled that "Markdown has no pages, so dropping the break itself is the right call", and that framing is right for a `w:type="page"` break — that is layout. A slide is different: it is a container in the source model, and every one of its shapes belongs to it. Losing the boundary loses document structure, not pagination.

This asks for no page number and no new concept. `Block::Rule` already exists, is already emitted (`src/shared/html.rs:432`, for `<hr>`), and already renders as `---` (`src/render/markdown/mod.rs:201`).

## Root cause

All three presentation parsers have the same shape — accumulate per slide, append, no separator:

- `src/formats/pptx/mod.rs` — `parse_shapes(sp_tree, &ctx, &mut blocks)?` writes straight into the shared `blocks`
- `src/formats/odf/mod.rs` — `parse_presentation` does `blocks.append(&mut title); blocks.append(&mut body);` per `draw:page`
- `src/formats/ppt/mod.rs` — `for (sid, blocks) in slides { out.extend(blocks); ... }`

## Scope

I have not scanned a corpus, so I will not claim a frequency. The structural argument is what I am confident in: any deck with a section divider, a full-bleed image slide, a quote slide, or a continuation slide has slides with no title placeholder, and those are the slides that merge.

Where every slide is titled, the current output is already correct and a separator only makes the boundary explicit.

## Proposed fix

Emit `Block::Rule` between slides — never leading, and never doubled by a slide that produces no blocks.

This is core-only. Every binding already maps `model::Block::Rule` (`node/src/document.rs:88`, `python/src/document.rs:83`, `wasm/src/document.rs:96`) and already declares the `rule` kind (`node/index.d.ts:60`, `python/anydoc/_anydoc.pyi:84`, `wasm/src/typescript.rs:47`), so no binding source or type surface changes and no new block variant.

PR follows.

Ready22Race:
Fixes #31.

## What

Emits `Block::Rule` between slides in the three presentation parsers (`pptx`, `ppt`, `odp`), which previously concatenated every slide's blocks with nothing in between.

A slide's title becomes a `Heading`, so decks that title every slide read correctly by accident. A slide with no title placeholder contributes no structural block, so its content is indistinguishable from a continuation of the previous slide — two untitled slides in a row produce two adjacent paragraphs, and bullet lists from different slides merge into one list.

## Why a thematic break

No new concept and no page number: `Block::Rule` already exists, is already emitted for `<hr>` (`src/shared/html.rs:432`), and already renders as `---` (`src/render/markdown/mod.rs:201`).

This is deliberately not the pagination #26 declined. That issue settled that a `w:type="page"` break is layout and dropping it is correct. A slide is a container in the source model rather than a layout artifact, so its boundary is document structure.

## Shape of the change

Each parser now accumulates one slide into a local `Vec<Block>` and appends it only if non-empty, pushing a `Rule` first when output already exists. That gives two guarantees worth stating:

- never leading — no `---` before the first slide
- never doubled — a slide that produces no blocks (empty layout, chrome-only placeholders) cannot leave a dangling or repeated break

`ppt` already accumulated per slide, so it only needed the guard.

## Scope: core only, no binding changes

`Block::Rule` already exists and every binding already handles it, so nothing outside `src/formats/` needs to move:

| binding | maps `model::Block::Rule` | declares the `rule` kind |
|---|---|---|
| node | `node/src/document.rs:88` | `node/index.d.ts:60` |
| python | `python/src/document.rs:83` | `python/anydoc/_anydoc.pyi:84` |
| wasm | `wasm/src/document.rs:96` | `wasm/src/typescript.rs:47` |

No new `Block` variant, no public API change, no type-surface change. Confirmed end to end through the Python binding on the deck from the issue:

```python
>>> anydoc.to_document(data).blocks
['heading', 'rule', 'paragraph', 'rule', 'paragraph']   # kinds
```

## Tests

Three unit tests in `src/formats/pptx/mod.rs`, building minimal packages in memory in the style of the `sheet` module's tests — they pin the behaviour rather than just the output:

- `untitled_slides_are_separated` — the defect itself: two title-less slides must not read as one
- `separator_never_leads` — a deck must not open with `---`, and a single slide yields only its own content
- `empty_slides_leave_no_dangling_separator` — a slide contributing no blocks must not add a break, neither doubled between two real slides nor trailing

## Verification

```
cargo fmt --all --check                                          clean
cargo clippy --workspace --all-targets --all-features -D warnings clean
cargo clippy -p anydoc-wasm --target wasm32-unknown-unknown       clean
cargo test --locked                                              194 passed, 0 failed
python -m unittest discover -s tests                             9 passed
```

Snapshot delta is 6 added `---` lines and 6 added blank lines across 6 snapshots, with **zero removals** — no content changed, only separators appeared:

```
tests/snapshots/snapshots__odp__pres.odp.snap                  | 2 ++
tests/snapshots/snapshots__ppt__handmade-multimaster.ppt.snap  | 2 ++
tests/snapshots/snapshots__ppt__handmade-sparsenotes.ppt.snap  | 2 ++
tests/snapshots/snapshots__ppt__pres.ppt.snap                  | 2 ++
tests/snapshots/snapshots__pptx__handmade-links.pptx.snap      | 2 ++
tests/snapshots/snapshots__pptx__pres.pptx.snap                | 2 ++
```

Also checked by hand against a deck whose slides 2 and 3 have no title placeholder (the case in the issue) and against `tests/fixtures/ppt/pres.ppt`.

## Note

Happy to take this in a different direction if you would rather the boundary be expressed some other way, or scope it to a subset of the three parsers. The behaviour change is visible in output, so it is your call whether it belongs in a minor bump.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/anydoc/pr/32"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788603077&installation_model_id=11126&pr_number=32&repository=firecrawl%2Fanydoc&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fanydoc%2Fpull%2F32&signature=d99110d5d2fb95185aad91b4885c4dc1794c50179f333da883e539a55e574de7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/formats/odf/mod.rs
+++ b/src/formats/odf/mod.rs
@@ -75,11 +75,19 @@ fn parse_presentation(pres: &Element, ctx: &Ctx) -> Result<Vec<Block>, ConvertEr
         let mut body = Vec::new();
         let mut notes = Vec::new();
         walk_shapes(page, ctx, &mut title, &mut body, &mut notes)?;
-        blocks.append(&mut title);
-        blocks.append(&mut body);
+        let mut page_blocks: Vec<Block> = Vec::new();
+        page_blocks.append(&mut title);
+        page_blocks.append(&mut body);
         // Speaker notes are included (fixed policy), set off as a quote.
         if !notes.is_empty() {
-            blocks.push(Block::BlockQuote(notes));
+            page_blocks.push(Block::BlockQuote(notes));
+        }
+        // Separate pages with a thematic break — see the pptx parser for why.
+        if !page_blocks.is_empty() {
+            if !blocks.is_empty() {
+                blocks.push(Block::Rule);
+            }
+            blocks.append(&mut page_blocks);
         }
     }
     Ok(blocks)
--- a/src/formats/ppt/mod.rs
+++ b/src/formats/ppt/mod.rs
@@ -343,6 +343,10 @@ impl Extractor {
         let mut used = vec![false; notes.len()];
         let mut out = Vec::new();
         for (sid, blocks) in slides {
+            // Separate slides with a thematic break — see the pptx parser.
+            if !blocks.is_empty() && !out.is_empty() {
+                out.push(Block::Rule);
+            }
             out.extend(blocks);
             for (i, (nid, nblocks)) in notes.iter_mut().enumerate() {
                 if !used[i] && sid.is_some() && *nid == sid {
--- a/src/formats/pptx/mod.rs
+++ b/src/formats/pptx/mod.rs
@@ -163,12 +163,23 @@ pub fn parse(bytes: &[u8]) -> Result<Document, ConvertError> {
             instance_counter: &instance_counter,
             slide_anchors: &slide_anchors,
         };
+        // A slide is a container, not a page: its content must not run into
+        // the next slide's. A thematic break is the structural separator the
+        // model already has, and it is emitted BETWEEN slides only — never
+        // leading, never doubled by an empty slide.
+        let mut slide_blocks: Vec<Block> = Vec::new();
         if targeted.contains(slide_path)
             && let Some(anchor) = slide_anchors.get(slide_path)
         {
-            blocks.push(Block::Paragraph(vec![Inline::Anchor(anchor.clone())]));
+            slide_blocks.push(Block::Paragraph(vec![Inline::Anchor(anchor.clone())]));
         }
-        parse_shapes(sp_tree, &ctx, &mut blocks)?;
+        parse_shapes(sp_tree, &ctx, &mut slide_blocks)?;
+        if !slide_blocks.is_empty() {
+            if !blocks.is_empty() {
+                blocks.push(Block::Rule);
+            }
+            blocks.append(&mut slide_blocks);
+        }
 
         // Speaker notes, set off as a quote (fixed policy: included). The
         // tree is loaded before the `if let` so the package borrow is not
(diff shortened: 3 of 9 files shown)


H0rowitz:
Independent verification, for whatever it's worth — we hit #31 on a Pandoc-authored deck while evaluating anydoc as the conversion engine for our ingestion pipeline, before finding the issue already filed.

Checked out `4b0edae` and ran it against that deck (3 slides; slide 2 has no title placeholder — exactly the pathological case from the issue):

- `cargo test --locked`: 194 passed, 0 failed, matching the PR's own verification block.
- The two separators appear at exactly the two slide boundaries, none leading, none doubled.
- A diff of the full converted output against `8eecca5` is 4 added lines (two `---`, two blanks), zero removals — and this is a producer (Pandoc) the fixture corpus doesn't currently cover, so it complements the PR's LibreOffice/handmade snapshot evidence rather than repeating it.

One note for anyone reading along, orthogonal to this PR: on that same deck the remaining structure loss is bullet formatting inherited from the layout (`<p:ph idx="1"/>` content placeholders lose their list markers) — the separate, bigger inheritance topic. The boundary fix makes such decks navigable regardless, and untitled slides stop merging into their predecessor.

blueagledev:
Independent verification. We're evaluating anydoc as the extraction layer for a documentation-ingestion pipeline that consumes the **document model** rather than the Markdown serializer, so slide boundaries are load-bearing for us. In fairness: we tested this PR and #95 against the same deck, and we've posted parallel notes there. Both work; the differences are below.

**Setup:** checked out from `refs/pull/32/head`, cargo/rustc 1.95.0. Test deck is Pandoc-authored: 2 slides, slide 1 titled with an image and speaker notes, **slide 2 has no title placeholder** — the pathological case from #31.

**`cargo test --locked`: 194 passed, 0 failed, 1 ignored.** (The lower count vs. `main`'s 217 is staleness, not skipped tests — this branch is based on `8eecca5`, roughly 23 commits behind `e754e1d`/v0.1.9. It still applies cleanly to current `main`.)

**Document model.** Baseline v0.1.9 gives 4 top-level blocks with no marker of any kind — `Heading -> Paragraph(Image) -> BlockQuote -> Table`. With this PR, 5 blocks:

```
Heading -> Paragraph(Image) -> BlockQuote -> Rule -> Table
```

The untitled slide 2 separates correctly, exactly as the PR describes, and the never-leading / never-doubled guarantees hold on our fixture.

**Where it left us short, and why we'd bet on #95.** Our output format numbers slides (`## Slide N`), so we need an ordinal, and it can't be recovered here by counting rules. This PR's own passing test `empty_slides_leave_no_dangling_separator` asserts that a 3-slide deck with an empty middle slide yields exactly one rule — correct behaviour for a separator, but it means a consumer counting rules labels the third slide "2". #95 carries the ordinal in the marker itself (`Inline::Anchor("slide-N")`, derived from `slide_paths.enumerate()`), so it survives empty and unparseable slides.

The secondary point is that `Block::Rule` is a content construct — indistinguishable from a genuine `<hr>` in source content — where an anchor is a dedicated identity. Your write-up argues a slide boundary is document structure rather than a layout artifact, and we agree; that argument seems to us to point at a structural marker rather than a content one.

**One thing this PR has that #95 does not:** ODF coverage. On `tests/fixtures/odp/pres.odp` this separates slides (11 blocks, one rule) while #95's output is byte-identical to base (10 blocks, no marker). If #95 is the one that lands, `.odp` needs the same treatment as a follow-up — that gap is real and it's yours.

The two conflict: this applies cleanly to `main`, after which #95 conflicts in 6 files (both edit the same statement), so whichever lands first forces the other to rebase.

None of this is a knock on the change — it does what it says and the reasoning in the description is sound. Posting the comparison because a maintainer choosing between two competing PRs is better served by it than by a thumbs-up on each.

PR #34, opened 2026-08-06
fix(rtf): treat an escaped CR or LF as \par

### LINKED ISSUE #33: rtf: `\` followed by a newline is dropped instead of read as \par, collapsing TextEdit-authored documents into one paragraph
## Summary

The RTF reader conventions define `\` immediately followed by a carriage return or line feed as equivalent to `\par` (RTF Specification 1.9.1, "Conventions of an RTF Reader"). The lexer instead emits that newline as an unknown control symbol, which the parser ignores — the paragraph mark vanishes.

This is the paragraph mark Cocoa's RTF writer uses: TextEdit and any `NSAttributedString` RTF export on macOS end essentially every body paragraph with `\` + LF rather than a spelled-out `\par`. So the failure is not an edge case: whole macOS-authored documents fuse into a single paragraph, and when the document contains a table, everything before the table is swallowed into the table's first cell (the table token scan misses the mark too).

Word- and Pandoc-authored RTF is unaffected — those writers spell out `\par`.

## Reproduction

Tested at v0.1.6 / HEAD `8eecca5`, Node binding and `cargo run --example convert` agree.

Minimal file — three paragraphs; each `\` below is the last byte of its line, i.e. `\`+LF:

```rtf
{\rtf1 Alpha\
Beta\
Gamma}
```

Observed:

```
AlphaBetaGamma
```

Expected:

```
Alpha

Beta

Gamma
```

Real-world: any document saved as RTF from TextEdit (or generated with `textutil -convert rtf`). Paragraphs, headings and lists all run together; with a table present, the document text before it lands inside the first cell.

## Root cause

`src/formats/rtf/lexer.rs` — `Lexer::control()` returns `Token::Symbol(b)` for every non-alphabetic byte after `\`, so `\r`/`\n` reach `control_symbol()` in `src/formats/rtf/mod.rs` (~line 491) and fall through its `_ => {}` arm.

One consequence worth stating for the fix's placement: `tables.rs` runs its own token scan, so handling the newline in the parser's symbol dispatch would fix body text but leave table cells broken. The token itself has to carry the paragraph mark.

## Proposed fix

Lex `\` + CR/LF as `Token::Word { name: "par" }`, so every consumer (body, notes, the table scan) shares the existing `\par` handling in one place. The LF of a `\<CR><LF>` pair is then plain-text whitespace the lexer already skips, so CRLF yields exactly one mark — and `\<LF>`, `\<CR><LF>`, `\<CR>` each break exactly one paragraph.

A few lines. PR follows with a Cocoa-shaped snapshot fixture.

H0rowitz:
Fixes #33.

## What

Lexes `\` immediately followed by CR or LF as the `par` control word, per the spec's reader conventions. This is the paragraph mark Cocoa's RTF writer (TextEdit, `NSAttributedString` exports) uses for essentially every body paragraph, so macOS-authored documents currently collapse into one paragraph — or into the first cell of a following table, since `tables.rs` runs its own token scan and misses the mark too.

## Shape of the change

The fix is in the lexer rather than the parser's symbol dispatch, deliberately: emitting `Token::Word { name: "par" }` gives every consumer — body text, notes, and the table scan — the existing `\par` handling in one place. A dispatcher-level fix would have left table cells broken.

The LF of a `\<CR><LF>` pair is then plain-text whitespace `next_token` already skips, so all three newline shapes yield exactly one mark each.

## Tests

- `escaped_newline_lexes_as_par` (lexer): `\<LF>`, `\<CR><LF>` and `\<CR>` each lex as one `par`; surrounding text bytes are untouched.
- `escaped_newline_ends_the_paragraph` (parser): the escaped form converts identically to the spelled-out `\par` form.
- New fixture `tests/fixtures/rtf/handmade-newline.rtf` + snapshot — Cocoa-shaped (`\cocoartf` prelude, color tables, escaped-LF and escaped-CRLF paragraph ends, a bold run, a `\'xx` escape), generated by `newline_rtf()` added to `tests/gen_fixtures.py`.

## Verification

```
cargo fmt --all --check                                            clean
cargo clippy --workspace --all-targets --all-features              clean
cargo test --locked                                                193 passed, 0 failed
```

(wasm32 clippy not run here — no wasm target installed locally.)

Snapshot delta: the new fixture's snapshot only; **zero existing snapshots changed**.

Minimal before/after (`{\rtf1 Alpha\<LF>Beta\<LF>Gamma}`):

```
before:  AlphaBetaGamma
after:   Alpha / Beta / Gamma as three paragraphs
```

Also checked against a real `textutil`-generated document with headings, nested lists and a table: before, the entire document text collapsed into the table's first cell; after, paragraphs, lists and the table come out as separate, correctly-typed blocks.

## Note

If you'd rather take this in your own direction, the issue carries everything needed to respec it — happy either way; the defect getting fixed is the point.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/anydoc/pr/34"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788613665&installation_model_id=11126&pr_number=34&repository=firecrawl%2Fanydoc&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fanydoc%2Fpull%2F34&signature=c74c226371eb9c5b4b991034ae8321d2186498d15657e9d870dfd8fe6e997ce0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/formats/rtf/lexer.rs
+++ b/src/formats/rtf/lexer.rs
@@ -60,6 +60,15 @@ impl<'a> Lexer<'a> {
                     _ => Some(Token::Byte(b'\'')),
                 };
             }
+            // An escaped CR or LF is a paragraph mark: the spec makes `\`
+            // followed by a newline equivalent to `\par` (Cocoa's writer
+            // ends every body paragraph this way). Lexing it as the control
+            // word keeps every consumer's `par` handling in one place; the
+            // LF of a `\<CR><LF>` pair is then plain-text whitespace, which
+            // `next_token` already skips.
+            if b == b'\r' || b == b'\n' {
+                return Some(Token::Word { name: "par", param: None });
+            }
             return Some(Token::Symbol(b));
         }
         let start = self.pos;
@@ -166,6 +175,23 @@ mod tests {
         }
         assert_eq!(bin, br"}}{\\");
         assert_eq!((opens, closes), (1, 1));
+    }
+
+    #[test]
+    fn escaped_newline_lexes_as_par() {
+        // `\<LF>`, `\<CR><LF>` and `\<CR>` are each one paragraph mark.
+        let src = b"{\\rtf1 Alpha\\\nBeta\\\r\nGamma\\\rDelta}";
+        let mut lexer = Lexer::new(src);
+        let (mut pars, mut text) = (0, Vec::new());
+        while let Some(t) = lexer.next_token() {
+            match t {
+                Token::Word { name: "par", .. } => pars += 1,
+                Token::Byte(b) => text.push(b),
+                _ => {}
+            }
+        }
+        assert_eq!(pars, 3);
+        assert_eq!(text, b"AlphaBetaGammaDelta");
     }
 
     #[test]
--- a/src/formats/rtf/mod.rs
+++ b/src/formats/rtf/mod.rs
@@ -1045,6 +1045,18 @@ mod tests {
     }
 
     #[test]
+    fn escaped_newline_ends_the_paragraph() {
+        // Cocoa's writer ends body paragraphs with `\` + newline instead of
+        // a spelled-out \par; both forms must break paragraphs identically.
+        let escaped = "{\\rtf1 Alpha\\\nBeta\\\r\nGamma}";
+        let spelled = r"{\rtf1 Alpha\par Beta\par Gamma}";
+        let convert =
+            |src: &str| crate::to_markdown_bytes(src.as_bytes(), crate::Format::Rtf).unwrap();
+        assert_eq!(convert(escaped), "Alpha\n\nBeta\n\nGamma\n");
+        assert_eq!(convert(escaped), convert(spelled));
+    }
+
+    #[test]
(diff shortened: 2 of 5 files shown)


tomsideguide:
This shipped in 0d24ad6 (v0.1.7), lexing a backslash before CR or LF as `par` in the lexer, same as you proposed. Closing as already landed, thanks for the writeup.

PR #36, opened 2026-08-06
fix(docx): map quote and code paragraph styles to their containers

### LINKED ISSUE #35: docx: Quote and Source Code paragraph styles flatten into plain paragraphs, losing the model's BlockQuote and CodeBlock
## Summary

Word marks quotations and code blocks with built-in paragraph *style names* rather than dedicated markup. The docx parser already resolves style names for structure — `heading N` / `Title` via `Styles::heading_level`, walking `basedOn` — but has no equivalent for quote and code styles, so styled documents flatten:

- a code block comes out as a plain paragraph, and its syntax-highlight character runs leak into Markdown as emphasis (`**fn** main() { … }`),
- a quotation loses the `>` container entirely.

Both target blocks already exist in the model (`Block::BlockQuote`, `Block::CodeBlock`), and the EPUB/HTML path already emits them (`src/shared/html.rs` for `<blockquote>`/`<pre>`), so the loss is specific to the docx frontend.

## Reproduction

Tested at v0.1.6 / HEAD `8eecca5`. Any Pandoc-authored docx shows both defects:

````
$ cat doc.md
> A quotation.

```rust
fn main() { println!("ok"); }
```

$ pandoc doc.md -o doc.docx   # then convert doc.docx
````

Observed:

```
A quotation.

**fn** main() { println!("ok"); }
```

Expected:

```
> A quotation.

```
fn main() { println!("ok"); }
```
```

The same flattening applies to documents styled in Word itself (built-in "Quote" / "Intense Quote") and to LibreOffice Writer exports ("Quotations", "Preformatted Text").

## Producer receipts

Style names actually emitted, verified against real files:

| producer | quote container | code container |
|---|---|---|
| Word built-ins | `Quote`, `Intense Quote` | `HTML Preformatted` |
| Pandoc writer | `Block Text` | `Source Code` |
| LibreOffice Writer | `Quotations` | `Preformatted Text` |

The one debatable name is `Block Text`: Word's own built-in of that name is a generic indented style, but it is what Pandoc's writer uses as its blockquote container, and Pandoc is probably the dominant producer of styled docx in document pipelines. Easy to drop from the set if you weigh that differently.

## Root cause

`src/formats/docx/content.rs::parse_paragraph` consults `heading_level` and numbering only; nothing maps a style name to a block container. `src/formats/docx/styles.rs` already has the exact mechanism needed (name matching walked through `basedOn`, as `heading_level` does).

## Proposed fix

A `Styles::para_role(id) -> Option<Quote|Code>` sibling of `heading_level`, matching the seven names above case-insensitively through `basedOn`, with these semantics:

- **Precedence**: heading and numbering win — a quote-styled paragraph carrying `outlineLvl` stays a heading, a numbered one stays a list item; the role only types otherwise-plain paragraphs.
- **Adjacency**: consecutive same-role paragraphs form one container (the source styles each paragraph; the container has no end marker). Word-authored multi-paragraph quotes and per-line code paragraphs read as one quote / one block.
- **Code is verbatim**: runs flatten to their text (killing the highlight-run emphasis leak), in-paragraph breaks become lines, no Markdown escaping inside the fence.

Core-only — no new `Block` variant, no binding or type-surface change. PR follows with a handmade fixture + snapshot.

Related but deliberately out of scope: character-level code spans (Pandoc's `Verbatim Char` rStyle) flatten to plain text too; that is a run-style topic for another day, this issue is only about the two block containers.

### PR DESCRIPTION
**Pull request by H0rowitz**
Fixes #35.

## What

Maps quote and code paragraph styles to the model's existing `BlockQuote` and `CodeBlock`, by style name through the `basedOn` chain — the same mechanism `heading_level` already uses. Names covered (case-insensitive): Word `Quote` / `Intense Quote` / `HTML Preformatted`, Pandoc `Source Code` / `Block Text`, LibreOffice `Quotations` / `Preformatted Text`.

## Semantics

- **Precedence** — heading and numbering win: a quote-styled paragraph carrying an outline level is still a heading, a numbered one is still a list item. The role only types otherwise-plain paragraphs (pinned by `heading_and_numbering_keep_precedence_over_style_roles`).
- **Adjacency** — consecutive same-role paragraphs form one container: the source styles each paragraph, the container itself has no end marker. Word multi-paragraph quotes become one quotation; per-line code paragraphs become one block, blank lines included.
- **Code is verbatim** — runs flatten to their text via `inlines_to_plain_text`, so syntax-highlight character runs stop leaking as `**fn**` emphasis; in-paragraph breaks become lines; fence content is unescaped (`|`, `*raw*` stay literal).
- Attachments anchored in a quote paragraph stay inside the quotation, as they do for list items.

## Scope: core only

`Block::BlockQuote` and `Block::CodeBlock { lang: None, .. }` already exist and every binding already maps them — no new variant, no binding source or type-surface change. `ParaRole` and `Styles::para_role` sit next to `heading_level` in `styles.rs`; `content.rs` gains two `ParaKind` variants and their `emit_paragraph` arms.

## Tests

- `para_role_matches_producer_names_through_basedon` (styles): direct names, a custom style inheriting its role via `basedOn`, and non-role names returning `None`.
- `quote_and_code_styles_become_their_blocks` (end-to-end): two quote paragraphs merge into one `BlockQuote`; a highlight-run code paragraph plus a second code paragraph become one verbatim `CodeBlock`.
- `heading_and_numbering_keep_precedence_over_style_roles` (end-to-end).
- New fixture `tests/fixtures/docx/handmade-styleroles.docx` + snapshot — adjacent quotes, highlight-run code with `|` and `*raw*` literals, a `basedOn`-derived quote style, a LibreOffice-named style — generated by `styleroles_docx()` added to `tests/gen_fixtures.py`.

## Verification

```
cargo fmt --all --check                                            clean
cargo clippy --workspace --all-targets --all-features              clean
cargo test --locked                                                194 passed, 0 failed
```

(wasm32 clippy not run here — no wasm target installed locally.)

Snapshot delta: the new fixture's snapshot only; **zero existing snapshots changed** — no current fixture uses these style names, which is also why the defect survived the corpus.

Real-file receipt — the code/quote section of a Pandoc-authored document, before:

```
**fn** main() { println!("héllo | monde"); }

Citation en bloc avec un [lien externe](https://…) et une note[^1].
```

after:

````
```
fn main() { println!("héllo | monde"); }
```

> Citation en bloc avec un [lien externe](https://…) et une note[^1].
````

(links and note references survive inside the quotation; the highlight-run emphasis is gone from the fence).

## Note

The set of names is the judgment call here — `Block Text` in particular is Pandoc's blockquote container but a generic indent style in Word's own catalog; trim it if you weigh that differently. If you'd rather reimplement along your own lines, the issue carries the full spec — happy either way.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/anydoc/pr/36"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788613683&installation_model_id=11126&pr_number=36&repository=firecrawl%2Fanydoc&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fanydoc%2Fpull%2F36&signature=9686cf07cf20d9bad852fea3eced042cabbd6a87835de9d357393e24e24d4e72"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


tomsideguide:
This shipped in 3e5066a (v0.1.7), mapping quote and code paragraph styles through the `basedOn` chain. Closing as already landed, thanks for the writeup.

Issue #43, opened 2026-08-06
(Optimization?) A 35 KB file is taking ~30s because each itemref re-parsing its part


AbhinavMir:
I was trying to see how anydoc holds up on malformed epubs and might've hit a wall.

  ```python
  import io, zipfile
  n = 6400
  buf = io.BytesIO()
  with zipfile.ZipFile(buf, "w", zipfile.ZIP_DEFLATED) as z:
      z.writestr("mimetype", "application/epub+zip", zipfile.ZIP_STORED)
      z.writestr("META-INF/container.xml",
          '<?xml version="1.0"?><container version="1.0" '
          'xmlns="urn:oasis:names:tc:opendocument:xmlns:container"><rootfiles>'
          '<rootfile full-path="c.opf" media-type="application/oebps-package+xml"/>'
          '</rootfiles></container>')
      items = "".join(f'<item id="i{i}" href="c.opf" media-type="application/xhtml+xml"/>' for
  i in range(n))
      refs = "".join(f'<itemref idref="i{i}"/>' for i in range(n))
      z.writestr("c.opf",
          '<?xml version="1.0"?><package xmlns="http://www.idpf.org/2007/opf" version="3.0">'
          f'<metadata/><manifest>{items}</manifest><spine>{refs}</spine></package>')
  open("quadratic.epub", "wb").write(buf.getvalue())
  ```

`cargo run --release --example convert -- quadratic.epub`

  35 KB, 27 seconds. At n=12800 (69 KB) it doesn't finish. 0.44s at n=800, 1.7s at 1600, 6.8s
  at 3200. Pointing the same itemrefs at a small chapter file instead of c.opf stays at 0.01s.

  src/formats/epub/mod.rs calls `optional_xml_part` once per href, and `Package::part` caches
  bytes but not the parsed tree, so a part referenced n times is parsed n times. The limits in
  src/package/limits.rs are all per-part or per-archive, so nothing bounds the total. `cargo
  fuzz run epub` finds it too. I'm happy to send a fix unless I was doing something wrong

  anydoc 0.1.6, release build, macOS arm64.

PR #44, opened 2026-08-06
Parse each package part once, and bound the total

AbhinavMir:
Packages now cache parsed part trees, so a part referenced many times parses once instead of once per reference, with a running node total bounding the parts too large to cache.

For issue #43.


abimaelmartell:
@cubic-dev-ai

PR #76, opened 2026-08-09
feat(math): preserve mathematical notation across every frontend

erkurtharun:
## Why

anydoc's own design note says the audience is AI agents and the output is
optimised for semantic quality. Mathematical notation was the one place that
promise broke, and it broke by changing values rather than presentation:

| The document says | anydoc gave |
|---|---|
| `10⁻³ mol/L` | `10-3 mol/L` |
| `H₂O` | `H2O` |
| `E = mc²` (MathML, no TeX annotation) | `E=mc2` |
| `⅓` as `<mfrac><mn>1</mn><mn>3</mn></mfrac>` | `13` |
| an equation in a slide or an ODF document | *nothing at all* |
| `&beta;` | `&amp;beta;` |

A lost bold is a lost presentation. A lost superscript is a different number,
and no consumer downstream can tell it was ever an exponent.

## What changed

**A run's position on the baseline** is now part of the model. `Style` gains
`vert_align`, read from `w:vertAlign` (DOCX), `style:text-position` (ODF), the
baseline percentage (PPTX and PPT), `sprmCIss` and `sprmCHpsPos` (DOC), and
`\super` / `\sub` / `\nosupersub` / `\up` / `\dn` (RTF). It renders as
`<sup>` / `<sub>`, which GFM already carries.

**Equations become LaTeX** in `$…$` / `$$…$$`. Two converters feed it:

- `shared/omml.rs` — Office Math, from DOCX and from the `a14:m` PowerPoint
  wraps in an `mc:AlternateContent`. That AlternateContent was skipped whole,
  so not even its fallback text survived.
- `shared/mathml.rs` — MathML, for HTML and EPUB, and for the formula
  sub-document ODF references through `draw:object`. Only some MathML carries
  a TeX annotation; without one the markup used to arrive as bare characters.

Structure is translated; **glyphs are not**. KaTeX takes Unicode operators and
Greek directly and has metrics for them, so `α` and `∑` are left as the
document wrote them — mapping them to commands would only be a chance to pick
the wrong one. The invisible operators (U+2061–2064) are the exception: they
carry no glyph, so passing them through would put unreadable codepoints in
front of a reader. An unmapped glyph is passed through, never guessed at.

**Two escaping defects** came out of this and are fixed. A paired `$` in
document text would open a math span and hand the rest of the document to the
Markdown parser; so would a bare `$` inside a TeX annotation. Both are now
escaped or refused.

**Named character references** were resolved for about forty names; everything
else fell through as literal text and was then escaped again, so `&beta;`
reached the reader as `&amp;beta;` — destroyed rather than merely unresolved.
The full HTML5 table is generated from the WHATWG `entities.json`. MathML
depends on these more than prose does.

**RTF character styles** (`\cs`) were not applied at all, so every property one
carried was lost — bold and italic as much as the script this branch went
after.

## Verification

- **260 unit/integration tests**, node bindings 14/14, `cargo fmt`,
  `cargo clippy -- -D warnings`, and clippy on `wasm32-unknown-unknown`.
- **A KaTeX gate.** The claim "inside the subset KaTeX implements" was never
  measured. `node/katex.test.mjs` walks the document model — which carries
  `latex` and `display` directly, so no re-parsing stands between what the
  converter produced and what is checked — and renders every equation with
  `throwOnError` and `strict: "error"`. Verified to fail: emitting an undefined
  control sequence breaks it.
- **Differential run against `main`** over all 72 fixtures: 66 byte-identical,
  6 changed, and all 6 carry mathematics. No panics.
- **Fuzzing.** 21,000 byte mutations, then 2,000 mutations of the math XML
  *inside* the packages. No panics. Rendering everything that survived found 9
  unrenderable equations, all from annotations passed through verbatim — that
  is the defect the annotation guard above fixes.
- **Adversarial corpus**: 20,000-deep nesting, a 120,000-cell table, 600,000
  entity references, a self-referencing ODF object, a cyclic RTF style. All
  bounded; the size and cell caps truncate while staying balanced. 20,005
  equations, zero KaTeX failures.
- **Performance**: on 20,000-paragraph documents, math-free conversion is
  unchanged (`+0.2%` DOCX, `−7.1%` EPUB, `−4.6%` RTF, byte-identical output).
  Math-heavy DOCX costs `+1.0%` while emitting twice the content.

## Known limits

- **Spreadsheets.** A cell's rich text still flattens (`10⁻³` → `10-3`). The
  loss is at the calamine boundary: `<si>` runs collapse into `Data::String`,
  and neither the shared-string table nor per-cell indices are public, so
  keying on anything else would be a guess. Preserving it means reading XLSX
  worksheets directly rather than through calamine — a separate decision.
- **Annotations with a misspelled control sequence** still pass through.
  Telling that from a macro would take a command table, and a table would
  reject legitimate LaTeX that is not in it.

## Related

PDFs bypass the document model and go through `pdf-inspector`, where the same
class of defect lives: scripts escape their line and reorder. That half is
firecrawl/pdf-inspector#326 — `10 -3 mol/L` → `10⁻³ mol/L`, and
`565.0 kg/m` + a stray `/kg` line → `565.0 kg/m³`. **PDF coverage here only
improves once that PR lands and a release follows**, since this crate pins
`pdf-inspector = "0.1.7"`. The chain was verified locally with a
`[patch.crates-io]` override, which is not part of this branch.


tomsideguide:
Thanks! Equations landed in v0.2.1 through #117 across docx, pptx, odt, epub and rtf, so closing this. Superscript and subscript runs outside equations are a separate question - worth their own issue if you want to pursue it.

PR #79, opened 2026-08-10
Ignore local .env files

ThomasGinter:
## Summary
- Add standard `.env` / `.env.*` ignore rules (with `.env.example` exceptions) so local secret files cannot be committed accidentally.
- Small hygiene fix; no runtime behavior change.

## Test plan
- [x] `git check-ignore .env` succeeds after change
- [ ] CI green


abimaelmartell:
this repo does not use env files, so not needed, thanks!

Issue #94, opened 2026-08-14
Expose slide boundaries for PPTX/PPT (the anchor already exists, it is just gated)


dksrudwns:
# Expose slide boundaries for PPTX/PPT (the anchor already exists, it is just gated)

## Summary

`to_document()` gives no way to tell where one slide ends and the next begins.
For a deck, `Document.blocks` is one flat sequence with nothing marking the
slide boundaries.

The information is already computed inside the PPTX reader — every slide gets a
`slide-N` anchor id — but the anchor node is only emitted for slides that some
internal hyperlink happens to target. Decks without slide-to-slide links (the
common case) get nothing.

Making that anchor unconditional, or exposing the slide index some other way,
would be a small change with a large payoff for downstream consumers.

## Why this matters

RAG / document-indexing pipelines chunk presentations **per slide**, so that a
retrieved passage can cite "slide 7". Without a boundary signal, a deck can only
be indexed as one undivided blob, or the boundary has to be guessed.

The obvious heuristic — treat each `heading` as a slide start — does not hold.
Slides that use plain text boxes instead of the title placeholder, and
table-only or image-only slides, produce no heading at all, so they get merged
into the preceding slide.

Measured on a 5-slide deck (titled / untitled-with-two-textboxes /
table-only / titled / title-only):

```
anydoc blocks:  heading, paragraph, paragraph, paragraph, table, heading, paragraph, heading
headings at:    [1, 6, 8]      -> 3 slides inferred, actual 5
```

## The mechanism already exists

`src/formats/pptx/mod.rs` (v0.1.9), lines 96–103:

```rust
// Every slide has a start anchor id so internal slide-to-slide links
// resolve after concatenation; the anchor node is emitted only on
// slides some link actually targets.
let slide_anchors: HashMap<String, String> = slide_paths
    .iter()
    .enumerate()
    .map(|(i, p)| (p.clone(), format!("slide-{}", i + 1)))
    .collect();
```

line 166:

```rust
if targeted.contains(slide_path)
    && let Some(anchor) = slide_anchors.get(slide_path)
{
    blocks.push(Block::Paragraph(vec![Inline::Anchor(anchor.clone())]));
}
```

The comment says every slide *has* a start anchor id; only the emission is
narrowed to link targets.

## Reproduction

Same three-slide deck, built twice — once with an internal slide-to-slide
hyperlink, once without:

```python
from pptx import Presentation
from pptx.util import Inches
import anydoc

def build(with_link, out):
[...]

PR #95, opened 2026-08-14
Expose slide boundary anchors for presentations

xianjianlf2:
## Summary
- emit `slide-N` anchors for every PPTX slide instead of only slides targeted by internal links
- add `slide-N` anchors for legacy PPT slides in presentation order
- cover PPTX with/without internal slide links and PPT slide ordering in document-model tests

## Tests
- `cargo fmt --check`
- `cargo test slide_anchor`
- `cargo test`
- `git diff --check`

Not passing locally: `cargo clippy --all-targets -- -D warnings` fails on existing unrelated `clippy::collapsible_match` warnings in `src/formats/rtf/tables.rs`.

Refs #94

---
## Summary by cubic
Expose slide boundary anchors for every PPTX and PPT slide and render them only for presentations. Previously PPTX anchors appeared only on slides targeted by internal links and unreferenced anchors rendered nothing; now each slide begins with a `slide-N` anchor, and Markdown emits `<a id="slide-N"></a>` for PPT/PPTX.

- **Details**
  - PPT/PPTX: Insert `Block::Paragraph` with `Inline::Anchor("slide-N")` before each slide; numbering follows presentation order and empty slides keep anchors. Removed targeted-slide detection in PPTX; dropped `SLIDE_REL` and `Relationships::iter`.
  - Markdown: Add `MarkdownOptions.render_unlinked_slide_anchors`; `to_markdown_bytes` enables it for `Format::Ppt`/`Format::Pptx`. Other formats still suppress unlinked anchors.
  - Tests: New unit tests for anchor presence/order and updated snapshots include leading slide anchors; please sanity‑check numbering.

<sup>Written for commit cb5a21e7daa659ff0ff15581bb282f0b23dfaba7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/anydoc/pull/95?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/formats/ppt/mod.rs
+++ b/src/formats/ppt/mod.rs
@@ -320,7 +320,8 @@ impl Extractor {
     fn end_segment(&mut self, id: Option<u32>) {
         self.flush_shape();
         flush_list(&mut self.current, &mut self.list_run);
-        if !self.current.is_empty() {
+        let keep_empty_slide = !self.current_is_notes && id.is_some();
+        if !self.current.is_empty() || keep_empty_slide {
             let blocks = std::mem::take(&mut self.current);
             self.segments.push((blocks, id, self.current_is_notes));
         }
@@ -633,5 +634,34 @@ impl Extractor {
                 self.current.push(Block::Paragraph(inlines));
             }
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn top_level_anchor_ids(blocks: &[Block]) -> Vec<&str> {
+        blocks
+            .iter()
+            .filter_map(|block| match block {
+                Block::Paragraph(inlines) => inlines.iter().find_map(|inline| match inline {
+                    Inline::Anchor(id) => Some(id.as_str()),
+                    _ => None,
+                }),
+                _ => None,
+            })
+            .collect()
+    }
+
+    #[test]
+    fn empty_slides_keep_anchor_positions() {
+        let mut extractor = Extractor::default();
+
+        extractor.end_segment(Some(10));
+        extractor.current.push(Block::Paragraph(vec![Inline::plain("Third slide")]));
+        extractor.end_segment(Some(30));
+
+        assert_eq!(top_level_anchor_ids(&extractor.into_blocks()), ["slide-1", "slide-2"]);
     }
 }
--- a/src/formats/ppt/mod.rs
+++ b/src/formats/ppt/mod.rs
@@ -342,7 +342,8 @@ impl Extractor {
         // slides), which order-based zipping would misattribute.
         let mut used = vec![false; notes.len()];
         let mut out = Vec::new();
-        for (sid, blocks) in slides {
+        for (slide_index, (sid, blocks)) in slides.into_iter().enumerate() {
+            out.push(Block::Paragraph(vec![Inline::Anchor(format!("slide-{}", slide_index + 1))]));
             out.extend(blocks);
             for (i, (nid, nblocks)) in notes.iter_mut().enumerate() {
                 if !used[i] && sid.is_some() && *nid == sid {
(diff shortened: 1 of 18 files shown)


blueagledev:
Independent verification. We're evaluating anydoc as the extraction layer for a documentation-ingestion pipeline that consumes the **document model** rather than the Markdown serializer, so slide boundaries are load-bearing for us. We tested this PR and #32 against the same deck; notes on both, and the clippy question answered.

**Setup:** merge-base `e754e1d` (v0.1.9), cargo/rustc/clippy 1.95.0, checked out from `refs/pull/95/head`. Test deck is Pandoc-authored: 2 slides, slide 1 titled with an image and speaker notes, **slide 2 has no title placeholder** — the pathological case from #31.

**`cargo test --locked`: 223 passed, 0 failed, 1 ignored.**

**On the clippy failure you flagged — your claim checks out, and it is not yours.** `cargo clippy --all-targets -- -D warnings` fails *identically on unmodified `main`*:

```
error: this `if` can be collapsed into the outer `match`
   --> src/formats/rtf/tables.rs:293:21
    = note: `-D clippy::collapsible-match` implied by `-D warnings`
error: could not compile `anydoc` (lib) due to 3 previous errors
```

Three `collapsible_match` at `src/formats/rtf/tables.rs` 293/302/422, on `main`, on this PR, and on #91 alike. Pre-existing and unrelated. Worth noting it's a newer-clippy artifact — CI pins `dtolnay/rust-toolchain@stable`, so it likely hasn't surfaced there yet.

**Document model on that deck.** Baseline v0.1.9 gives 4 top-level blocks with no marker of any kind — `Heading -> Paragraph(Image) -> BlockQuote -> Table` — so slide 2's table abuts slide 1 with nothing between. With this PR, 6 blocks:

```
Paragraph[Anchor("slide-1")] -> Heading -> Paragraph(Image) -> BlockQuote
  -> Paragraph[Anchor("slide-2")] -> Table
```

The untitled slide separates correctly, and the marker carries the ordinal.

**Why we'd prefer this over #32, having run both.** #32 works — it emits `Block::Rule` and separates the same untitled slide correctly — but it gives a *boundary* where this PR gives an *identity*. Two consequences for a model consumer:

- Ordinals can't be recovered from #32 by counting rules. Its own (passing) test `empty_slides_leave_no_dangling_separator` asserts that a 3-slide deck with an empty middle slide yields exactly one rule, so the third slide would be labelled "2". This PR derives ids from `slide_paths.enumerate()`, so numbering survives empty and unparseable slides — `empty_slides_keep_anchor_positions` covers it.
- `Block::Rule` is a content construct: indistinguishable from a genuine `<hr>` in source content.

It's also the smaller change than the diffstat suggests — v0.1.9 already builds `slide-{n}` ids and already pushes `Block::Paragraph(vec![Inline::Anchor(..)])`, gated on `targeted.contains(..)`. This mostly deletes that gate. The `slide-N` convention is already upstream's own.

**One gap, in fairness to #32:** this PR doesn't touch the ODF frontend. On `tests/fixtures/odp/pres.odp` its output is byte-identical to base (10 blocks, no marker), where #32 does separate `.odp` (11 blocks, one rule). If `.odp` parity is wanted, that's a follow-up here.

The two PRs conflict — #32 applies cleanly to `main`, after which this one conflicts in 6 files (both edit the same statement), so whichever lands first forces the other to rebase.

Happy to re-run any of this on other fixtures if it's useful.

PR #108, opened 2026-08-19
fix: escape code span pipes in table cells

### LINKED ISSUE #84: markdown: a pipe inside a code span in a table cell is not escaped, splitting the GFM row and dropping later columns
`|` is escaped in table cells for plain text, emphasis, and URLs, but not for code spans. `push_code_span()` never receives the `InlineContext`, so it emits raw pipes, which split the GFM table row and silently drop every column to the right of the code cell.

## Expected

The GFM spec (§4.10 Tables extension, [Example 200](https://github.github.com/gfm/#example-200)) is explicit that this applies inside inline spans:

> Include a pipe in a cell's content by escaping it, **including inside other inline spans**:
>
> ```
> | b `\|` az |
> ```
> → `<td>b <code>|</code> az</td>`

The repo already treats this as settled for the other three paths — `escape.rs:82` escapes `'|' if ctx == InlineContext::TableCell` for plain and styled text, and `format_url` percent-encodes it under a comment that states the rule outright:

```rust
// src/render/markdown/escape.rs:151-152
// Raw pipes split GFM table cells.
'|' => escaped.push_str("%7C"),
```

with the existing test `url_pipes_cannot_split_table_cells` (`tests.rs:289`) and the plain-text case in `table_basic` (`tests.rs:232`, `Ann | Bob` → `Ann \| Bob`). Only the code-span path is missing.

## Actual

Minimal `.epub` with a two-column table:

```html
<table>
<thead><tr><th>Operator</th><th>Meaning</th></tr></thead>
<tbody>
<tr><td><code>a | b</code></td><td>bitwise or</td></tr>
<tr><td><pre>ls | wc -l</pre></td><td>shell pipeline</td></tr>
<tr><td>plain a | b</td><td>escaped correctly</td></tr>
</tbody>
</table>
```

`cargo run --example convert -- pipes.epub` on v0.1.8 (`4e3089b`):

```
| Operator | Meaning |
| --- | --- |
| `a | b` | bitwise or |
| `ls | wc -l` | shell pipeline |
| plain a \| b | escaped correctly |
```

The delimiter row declares two columns, but the first two body rows emit three cells each. GFM ignores cells past the header width, so feeding that output to a GFM renderer (marked 18, `gfm: true`) loses the `Meaning` value and tears the code span in half:

```html
<tr><th>Operator</th><th>Meaning</th></tr>
<tr><td>`a</td><td>b`</td></tr>                          <!-- "bitwise or" gone -->
<tr><td>`ls</td><td>wc -l`</td></tr>                     <!-- "shell pipeline" gone -->
<tr><td>plain a | b</td><td>escaped correctly</td></tr>  <!-- correct -->
```

Expected first body row is `| `a \| b` | bitwise or |`, which renders as `<td><code>a | b</code></td><td>bitwise or</td>` with both columns intact.

## Failing tests

In the style of the existing `url_pipes_cannot_split_table_cells`:

```rust
#[test]
fn code_span_pipes_cannot_split_table_cells() {
    let md = doc(vec![table_from(
        vec![
            vec![
                Cell::from_inlines(vec![Inline::plain("Operator")]),
                Cell::from_inlines(vec![Inline::plain("Meaning")]),
            ],
            vec![
                Cell::from_inlines(vec![styled("a | b", Style { code: true, ..Style::PLAIN })]),
                Cell::from_inlines(vec![Inline::plain("bitwise or")]),
            ],
        ],
        1,
    )]);
    assert_eq!(md, "| Operator | Meaning |\n| --- | --- |\n| `a \\| b` | bitwise or |\n");
}

#[test]
fn code_block_pipes_cannot_split_table_cells() {
    let cell = Cell::new(vec![Block::CodeBlock { lang: None, text: "ls | wc -l".into() }]);
    let md = doc(vec![table_from(vec![vec![cell]], 0)]);
    assert_eq!(md, "|  |\n| --- |\n| `ls \\| wc -l` |\n");
}
```

Both fail on `4e3089b`, while `url_pipes_cannot_split_table_cells` passes:

```
test render::markdown::tests::url_pipes_cannot_split_table_cells ... ok
test render::markdown::tests::code_block_pipes_cannot_split_table_cells ... FAILED
test render::markdown::tests::code_span_pipes_cannot_split_table_cells ... FAILED

  left: "| Operator | Meaning |\n| --- | --- |\n| `a | b` | bitwise or |\n"
 right: "| Operator | Meaning |\n| --- | --- |\n| `a \| b` | bitwise or |\n"
```

## Root cause

Pipe escaping lives in `escape_text()`, keyed on `InlineContext::TableCell` (`escape.rs:82`). Code spans are a bypass around it: `inline.rs:207` routes code-styled runs to `push_code_span()`, which takes no `InlineContext` and only handles newlines and the backtick fence. `table.rs:169` sends cell-level `Block::CodeBlock` down the same function. Because GFM splits table rows *before* inline parsing, backticks do not protect a pipe — the row is already cut by the time the code span would exist.

## Affected inputs

- **EPUB** — `<code>`, `<kbd>`, `<samp>`, `<tt>` (`shared/html.rs:690`) and `<pre>` (`shared/html.rs:423`).
- **docx, rtf, odt, ods, odp** — the shared `Source Code` / `HTML Preformatted` / `Preformatted Text` paragraph styles (`shared/blockstyle.rs:22`). The existing `rtf/table.rs:219` and `odf/mod.rs:275` tests both assert a `CodeBlock` landing inside a table cell, so this path is reachable there.
- **CSV and XLSX** are unaffected — they only emit `Inline::plain`.

Trigger conditions are ordinary for technical documents: operator tables (`a | b`), shell pipelines (`ls | wc -l`), union types (`string | null`), regex alternation, and Markdown table examples themselves. The snapshot corpus happens to contain no table cell with code *and* a pipe, which is why CI stayed green.

Severity feels moderate rather than cosmetic: for a converter positioned as producing LLM-ready Markdown, the failure mode is silent loss of a neighbouring column rather than visibly broken output.

## Suggested fix

Give `push_code_span` the context and escape pipes in table cells (two call sites: `inline.rs:208`, `table.rs:169`):

```diff
-pub(crate) fn push_code_span(text: &str, out: &mut String) {
+pub(crate) fn push_code_span(text: &str, ctx: InlineContext, out: &mut String) {
     let text = text.replace('\n', " ");
+    // A raw pipe splits a GFM table cell even inside a code span.
+    let text = if ctx == InlineContext::TableCell { text.replace('|', "\\|") } else { text };
     let fence = backtick_fence(&text, 1);
```

With that patch `cargo test` is fully green locally (210 unit tests including the two above, plus the integration and snapshot suites) with no snapshot changes. Happy to open a PR if that's welcome.

---

This report was produced with AI assistance; I reproduced it locally and reviewed every conclusion myself.

CooperSheroy:
## Summary
- escape pipes inside inline code spans rendered in GFM table cells
- use the same table-cell protection for code blocks flattened into table cells
- add focused renderer regressions for both inline code and code blocks

Fixes #84.

## Validation
- cargo fmt --all --check
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test --locked
- git diff --check

---
## Summary by cubic
Escapes pipes in inline code and in code blocks rendered inside GFM table cells so they no longer split cells. Previously we emitted raw pipes in these contexts; now we render them as \| while preserving backtick fencing. Non-table contexts are unchanged.

**Notes for review**
- Inline renderer: in table-cell context, route code spans through a new code-path that escapes '|' before choosing the backtick fence.
- Table renderer: apply the same escape when flattening code blocks into table cells.
- Tests: add focused regressions for inline code and code blocks to confirm pipes cannot split table cells.

<sup>Written for commit db419ab0eadf5b607334f3710d098ddc123fb6a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/anydoc/pull/108?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/render/markdown/inline.rs
+++ b/src/render/markdown/inline.rs
@@ -205,7 +205,11 @@ fn render_text_run(
     }
     if !core.is_empty() {
         if style.code {
-            push_code_span(core, out);
+            if ctx == InlineContext::TableCell {
+                push_table_cell_code_span(core, out);
+            } else {
+                push_code_span(core, out);
+            }
         } else {
             let mut open = String::new();
             if style.strike {
@@ -233,7 +237,16 @@ fn render_text_run(
 }
 
 pub(crate) fn push_code_span(text: &str, out: &mut String) {
+    push_code_span_inner(text, false, out);
+}
+
+pub(crate) fn push_table_cell_code_span(text: &str, out: &mut String) {
+    push_code_span_inner(text, true, out);
+}
+
+fn push_code_span_inner(text: &str, escape_table_pipes: bool, out: &mut String) {
     let text = text.replace('\n', " ");
+    let text = if escape_table_pipes { text.replace('|', "\\|") } else { text };
     let fence = backtick_fence(&text, 1);
     let pad = if text.starts_with('`') || text.ends_with('`') { " " } else { "" };
     let _ = write!(out, "{fence}{pad}{text}{pad}{fence}");
--- a/src/render/markdown/table.rs
+++ b/src/render/markdown/table.rs
@@ -166,7 +166,7 @@ fn cell_block_text(block: &Block, rc: &Ctx, parts: &mut Vec<String>) {
             let t = text.trim();
             if !t.is_empty() {
                 let mut s = String::new();
-                crate::render::markdown::inline::push_code_span(t, &mut s);
+                crate::render::markdown::inline::push_table_cell_code_span(t, &mut s);
                 parts.push(s);
             }
         }
--- a/src/render/markdown/tests.rs
+++ b/src/render/markdown/tests.rs
@@ -296,6 +296,23 @@ fn url_pipes_cannot_split_table_cells() {
 }
 
 #[test]
+fn code_span_pipes_cannot_split_table_cells() {
+    let cell =
+        Cell::from_inlines(vec![styled("left | right", Style { code: true, ..Style::PLAIN })]);
+    let md =
+        doc(vec![table_from(vec![vec![cell, Cell::from_inlines(vec![Inline::plain("x")])]], 0)]);
+    assert_eq!(md, "|  |  |\n| --- | --- |\n| `left \\| right` | x |\n");
+}
+
+#[test]
+fn code_block_pipes_cannot_split_table_cells() {
+    let cell = Cell::new(vec![Block::CodeBlock { text: "a | b".into(), lang: None }]);
+    let md =
+        doc(vec![table_from(vec![vec![cell, Cell::from_inlines(vec![Inline::plain("x")])]], 0)]);
+    assert_eq!(md, "|  |  |\n| --- | --- |\n| `a \\| b` | x |\n");
+}
+
+#[test]
 fn url_angle_brackets_are_encoded_without_bracketing() {
     let md = doc(vec![Block::Paragraph(vec![Inline::Link {
         content: vec![Inline::plain("link")],


tomsideguide:
Fixed in aa6ab89, which also covers a pre block in a cell and the backslash run case. Thanks!

PR #112, opened 2026-08-19
feat(sheet): read every Excel container in-house and drop calamine

### LINKED ISSUE #8: XLSX merged-cell spans are clipped to the populated range

### LINKED ISSUE #9: XLSX conversion silently treats hidden rows and columns as visible
## Summary

XLSX cells in hidden rows and columns are returned by `to_document()` and rendered by `to_markdown()` exactly like visible cells, with no visibility metadata or other indication that they are hidden in the source workbook.

This can make calculation/template content appear user-visible and authoritative in LLM input even though it is not visible when the workbook is opened normally.

## Reproduction

```python
from pathlib import Path

import anydoc
from openpyxl import Workbook

path = Path("hidden-content.xlsx")

workbook = Workbook()
sheet = workbook.active
sheet.title = "Visibility Example"
sheet["A1"] = "Visible row"
sheet["A2"] = "Hidden row"
sheet.row_dimensions[2].hidden = True
sheet["B1"] = "Hidden column"
sheet.column_dimensions["B"].hidden = True
workbook.save(path)

print([name for name in dir(table) if not name.startswith("_")])
```

## Observed

```text
|  |  |
| --- | --- |
| Visible row | Hidden column |
| Hidden row |  |

['grid', 'header_rows', 'kind']
```

The hidden row and hidden column are indistinguishable from visible content in both outputs.

## Expected

Visibility should not be silently flattened away.

For Markdown, a reasonable default would be to omit hidden rows and columns, with an option to include them when desired. For `to_document()`, retaining row/column visibility metadata would let callers make that choice themselves.

If including hidden content unconditionally is intentional, the behavior should at least be documented because it materially affects what downstream LLMs see.

### LINKED ISSUE #27: Spreadsheet number formats are dropped, rendering 7.5% as 0.075
## Summary

Spreadsheet cells carrying a `numFmt` are rendered from their stored value with the display format dropped. A cell that reads `7.5%` in Excel converts to `0.075`, and one that reads `$1,234.50` converts to `1234.5`.

The percent case is the damaging one: it is not a cosmetic loss but a value that is wrong by two orders of magnitude for any consumer reading the Markdown, which for this library is usually an LLM. A conversion-rate column of `0.075` reads as 0.075%, not 7.5%, and nothing in the output signals that a format was dropped.

Dates are handled correctly, which makes the behaviour inconsistent: `mm/dd/yyyy` renders as `2026-03-15`, so a reader could reasonably assume other display formats survive too.

## Reproduction

Tested with `firecrawl-anydoc==0.1.6` and `openpyxl==3.1.5`.

```python
import datetime
from pathlib import Path

import anydoc
from openpyxl import Workbook

path = Path("number-formats.xlsx")
workbook = Workbook()
sheet = workbook.active
sheet.title = "Formats"
sheet.append(["kind", "value"])
sheet.append(["percent", 0.075])
sheet.append(["currency", 1234.5])
sheet.append(["date", datetime.date(2026, 3, 15)])
sheet["B2"].number_format = "0.0%"
sheet["B3"].number_format = '"$"#,##0.00'
sheet["B4"].number_format = "mm/dd/yyyy"
workbook.save(path)

print(anydoc.to_markdown(str(path)))
```

## Actual

```
| kind | value |
| --- | --- |
| percent | 0.075 |
| currency | 1234.5 |
| date | 2026-03-15 |
```

## Expected

The displayed value, as the workbook shows it:

```
| kind | value |
| --- | --- |
| percent | 7.5% |
| currency | $1,234.50 |
| date | 2026-03-15 |
```

## Cause

`format_data` in `src/formats/sheet/mod.rs` switches on calamine's `Data` enum alone. Dates survive because calamine resolves them to `Data::DateTime` itself; percent and currency cells arrive as `Data::Float` and fall through to `format_float`, which never consults the cell's number format.

calamine 0.36.1 does not appear to expose the `numFmt` code on its public API, so this likely needs `xl/styles.xml` read directly — resolving `cellXfs[xfId].numFmtId` against the built-in format IDs plus any `numFmts` overrides — which the existing package/XML layer already supports. The built-in IDs alone (9 and 10 for percent, 5-8 and 37-44 for currency and accounting) would cover most real workbooks.

If full format emulation is out of scope, even handling percent alone would remove the wrong-by-100x failure, which is the part that silently corrupts meaning rather than just dropping presentation.

## Context

Found while benchmarking anydoc for accuracy across a mixed corpus of real business documents. Text recall was 100% on every file tested and the library was 15-40x faster than the alternative I compared against, so this stood out as the one case where output was confidently wrong rather than merely lossy.

tomsideguide:
Number formats were dropped, so a cell showing `7.5%` converted to `0.075`. calamine does not expose the format code, so this needed the styles part read directly, which meant reading the workbook directly.

All four containers are now read in-house: xlsx and xlsm from XML, xlsb from its binary records, xls from BIFF in an OLE container. They share one format engine and one grid assembly, so the same workbook saved in any of them converts identically. The `catch_unwind` barrier is gone with calamine; malformed input raises typed errors.

Hidden rows, columns and sheets are omitted now, and a merge extending past the populated range keeps its full extent.

New cap: a sheet grid spans its populated extent, so two cells at opposite corners described a whole sheet of empty positions from a few hundred bytes. `MAX_GRID_SLOTS` charges that before materializing. A fully populated legacy .xls at its maximum 65536x256 exceeds it and now returns `ResourceLimit`.

Fuzzing reached almost none of this before: random bytes never form an OLE container or a zip holding `workbook.bin`. Three targets wrap their input in a valid container so mutation reaches the record and format-code parsers, with checked-in seeds.

Closes #27
Closes #8
Closes #9

---
## Summary by cubic
Reads every Excel container in-house and removes `calamine`, restoring correct number-format rendering and identical output across `.xlsx`, `.xlsm`, `.xlsb`, and `.xls`. Previously percentage/currency formats were lost; now parsed format codes render them, empty codes fall back to General, malformed input raises typed errors, and only date/time parts render with ISO-like output.

- Omits hidden rows, columns, and sheets; preserves merge extent beyond populated cells. `GridBuilder::keep_covered_tail` keeps merged tails during trimming.
- Caps materialized grid slots across the workbook with `MAX_GRID_SLOTS` (4,000,000); a fully populated legacy `.xls` at 65536×256 now returns a ResourceLimit error.
- Resolves the workbook’s main part before conventional paths and requires a resolvable workbook; also requires a workbook when only the conventional binary part exists. Opens the package once, classifies the main part from cached bytes, picks the reader from the part’s bytes (XML vs binary), treats a leading BOM as XML, and propagates limit errors.
- Number-format engine: matches AM/PM tokens over chars; treats elapsed brackets as duration-only; keeps literals around General and keeps a trailing decimal point with no placeholder; keeps the time part when a combined date–time format is applied to a sub‑day serial; finds the closest fraction by walking the exact ratio’s continued fraction with a saturated denominator bound; drops the clock for date‑only formats; keeps the fictitious 1900‑02‑29 distinct.
- Truncated `.xls` BIFF streams report a structured error with a warning instead of panicking.
- Adds fuzz targets for the binary readers and the format engine; the `xlsb` target reaches cell, style, and shared-string readers, and seeds cover AM/PM token paths while preserving whitespace in format codes.

<sup>Written for commit 3e66203137caed17839a0b42b897999e6cb5872f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/anydoc/pull/112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


delong3:
Nicely done!

PR #117, opened 2026-08-20
feat(math): convert equations to latex math

tomsideguide:
Equations were dropped in docx, pptx, odt and rtf, and epub leaked MathML token text plus the TeX annotation mashed together.

Adds `Inline::Math` / `Block::Math` carrying LaTeX, rendered as GFM `$...$` / `$$` blocks, with a shared OMML converter (docx, pptx via `a14:m`, rtf via its math destinations built into the same element tree) and a MathML converter (epub, odt formula objects). Plain-text `$` is now escaped so it can never pair with emitted delimiters.

Watch: the four spreadsheet snapshots change only by `$` -> `\$`; rtf math was verified against hand-built Word-shaped input, not a real Word file.

---
## Summary by cubic
Converts equations to LaTeX and renders them as GitHub-flavored math. Before, OMML/MathML were dropped or leaked tokens; now inline math uses $...$, block math uses $$, and non-math $ is escaped to prevent accidental math.

- Introduces `Inline::Math` and `Block::Math` carrying LaTeX (without delimiters). Markdown renders inline with $...$ and blocks with $$; in tables, block math is coerced to inline and pipes are escaped, preserving existing escapes and zero‑width spaces. Plain‑text flattening includes formula source.
- Converts OMML in `.docx`/`.pptx` (including `a14:m`) and RTF math destinations, and MathML in `.odt`/EPUB via shared converters in `src/shared/math`; adds tests and documents math support. Hardens delimiters: escapes bare `$` in text and `$` inside block formulas; fixes RTF run properties inside math zones.

**Migration**
- Node, Python, and WASM APIs add a `math` variant to block and inline kinds, with `text` holding the LaTeX source. Action: update enum matches to handle or ignore `math`; if you render Markdown, ensure `$`/`$$` delimiters are acceptable.

<sup>Written for commit a5556bbceecadcc8a5d54f871e4196ed0e17eca4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/anydoc/pull/117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tomsideguide:
@cubic-dev-ai

PR #129, opened 2026-08-23
fix(docx): continue counters across numIds that share an abstract (#96)

### LINKED ISSUE #96: DOCM to MD Conversion - Headings to not match expected numbering from document.
When running my DOCM through anydoc for a markdown output, I commonly get heading/subheading numberings that do not match the input DOCX file. Is there any way to change this or diagnose WHY is is doing this?  

Additional Information 
- ToC is correct
- Multiple section '1' at the start of the document.  
- In 'Section 4' it went in this order 
    > Section 4
    > Section 4.1
    > Section 1.1
    > Section 4.2

EDIT: Applying a new style and reapplying the intended style fixes the issues. If needed I could try to add a pre/post processing script to check for this, but I need to know what anydoc is interpreting as incorrect. 

*I can not attach the document directly as it contains proprietary company information.

Mr-Neutr0n:
Fixes #96

## Summary

Heading and list numbering now continues across `w:num` instances that share one `w:abstractNum`, instead of restarting at every instance switch.

## Root cause

Counters were keyed per numId. Word documents (and converter output) routinely split one logical outline across several numIds of the same abstract, so every direct-`numPr` vs style-inherited switch reset the sequence: the reported file renders "Section 4 → 4.1 → 1.1 → 4.2" with multiple "Section 1"s. Reapplying the style "fixes" it only because it removes the direct numPr, collapsing everything to a single numId.

Reproduced with a minimal OOXML fixture: two numIds sharing abstract 0 produce Section 1, Section 2, **Section 1**, **1.1** on this branch's base commit.

## What changed

`Counters` keys state by the resolved abstract definition (after numStyleLink indirection). Entering through an instance carrying `lvlOverride` restarts exactly its overridden levels; plain instances continue the shared sequence. `startOverride` starts are honored as before.

Evidence this matches Word: the real-world `text.docx` fixture has consecutive items on numId 5 then numId 4 whose source text labels them "Roman starting at four", "Roman five" - previously rendered IV then **I**; now IV then V.

## Snapshot updates (intentional, flagged)

- handmade-numbering: "Two-one independent counter" now continues (5.) rather than restarting (1.), and the style-referenced paragraph follows the merged count (8.). Override cases (10., 7)) unchanged. If bare shared instances were meant to stay independent despite Word behavior, that contract can be restored behind an option -- but parity with Word is what #96 reports as missing.
- text.docx / malformed variants: "I." becomes "V.", matching the document's own labels.

## Verification

- New regression test `instances_sharing_an_abstract_continue_one_sequence` (continues across n1→n2, restarts at startOverride=10).
- Full suite: 287 passed incl. snapshots; clippy clean; fmt clean.

This change was prepared with AI assistance under human direction and review.

---
## Summary by cubic
Continue DOCX list and heading counters across `w:num` instances that resolve to the same `w:abstractNum`, matching Word. Previously counters keyed by `numId` restarted on each instance switch; now they key by the resolved abstract and restart only levels overridden by `w:lvlOverride`/`w:startOverride`.

- Key counters by abstract identity (after `w:numStyleLink` resolution). Track per-instance overridden levels and restart them on first use under that instance; plain instances continue the shared sequence.
- Store the last active instance as `Option<u64>` to avoid sentinel collisions with legal max `numId` values.
- Add regression tests `instances_sharing_an_abstract_continue_one_sequence` and `override_restart_does_not_leak_into_the_next_instance`. Update snapshots where shared instances now continue (e.g., "Two-one independent counter" -> 5., "I." -> "V."); override-driven restarts are unchanged.

<sup>Written for commit 52dfa5ebb8082a3343735793f409af24ae914abe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/anydoc/pull/129?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/formats/docx/mod.rs
+++ b/src/formats/docx/mod.rs
@@ -192,6 +192,47 @@ mod tests {
     }
 
     #[test]
+    fn instances_sharing_an_abstract_continue_one_sequence() {
+        // #96: Word treats every w:num that references one w:abstractNum as
+        // the same logical list; converters emit exactly this shape for a
+        // single outline split across several numIds. Counters must continue
+        // across the instance switches, and a startOverride must restart.
+        let document = r#"<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
+            <w:body>
+            <w:p><w:pPr><w:numPr><w:ilvl w:val="0"/><w:numId w:val="1"/></w:numPr></w:pPr><w:r><w:t>one</w:t></w:r></w:p>
+            <w:p><w:pPr><w:numPr><w:ilvl w:val="0"/><w:numId w:val="1"/></w:numPr></w:pPr><w:r><w:t>two</w:t></w:r></w:p>
+            <w:p><w:pPr><w:numPr><w:ilvl w:val="0"/><w:numId w:val="2"/></w:numPr></w:pPr><w:r><w:t>three</w:t></w:r></w:p>
+            <w:p><w:pPr><w:numPr><w:ilvl w:val="0"/><w:numId w:val="3"/></w:numPr></w:pPr><w:r><w:t>ten</w:t></w:r></w:p>
+            </w:body></w:document>"#;
+        let numbering = r#"<w:numbering xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
+            <w:abstractNum w:abstractNumId="7">
+                <w:lvl w:ilvl="0"><w:numFmt w:val="decimal"/><w:start w:val="1"/></w:lvl>
+            </w:abstractNum>
+            <w:num w:numId="1"><w:abstractNumId w:val="7"/></w:num>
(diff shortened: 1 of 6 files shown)


Dark1886:
Used this branch against the original document that lead to issue #96, headings appear to be numbered correctly now.

Mr-Neutr0n:
Hi @tomsideguide - a gentle nudge on this docx counter fix; the branch still applies and the fix was confirmed against the original document by @Dark1886. Happy to update anything, and fine to close if not wanted. Thank you.

PR #130, opened 2026-08-23
feat: decrypt password-protected OOXML when a password is supplied (#102)

### LINKED ISSUE #102: Accept a password for encrypted OOXML files instead of terminating at `Encrypted`
## Summary

A password-protected `.xlsx`/`.docx`/`.pptx` always ends in `ConvertError::Encrypted`, and there is no way to supply the password. The rejection is precise and deliberate today: `probe_ole` (`src/package/archive.rs`) identifies the OLE container with `EncryptionInfo`/`EncryptedPackage` streams and returns `Encrypted`.

For a caller that *has* the password, the file is still unconvertible.

## Why this matters

Document-ingestion pipelines (the LLM/RAG use case the README targets) receive password-protected spreadsheets and documents routinely, and the password is usually known: it arrived alongside the file.

Rust users can already work around this by composing the `office-crypto` crate with `to_markdown_bytes`:

```rust
let decrypted = office_crypto::decrypt_from_bytes(std::fs::read(path)?, "secret")?;
let md = anydoc::to_markdown_bytes(&decrypted, None)?;
```

But Node, Python, wasm, and CLI users have no equivalent: there is no solid OOXML-decryption library on those sides, so for the bindings the file is a dead end.

## Proposed shape (narrowest version)

- An optional password on the byte-level entry points, e.g. `to_markdown_bytes(bytes, format, password: Option<&str>)` or `_with_password` variants, mirrored in the bindings.
- CLI: `--password <pw>` (plus perhaps an env-var fallback, since argv leaks into shell history and `ps`).
- Where `probe_ole` currently returns `Encrypted`: if a password was supplied, decrypt via `office-crypto` and re-enter `Format::from_bytes` -> conversion with the plaintext. All existing resource limits then apply to the decrypted package unchanged. No password supplied, or wrong password: `Encrypted`, exactly as today.

`office-crypto` (v0.3.0) is pure Rust, its dependencies are all RustCrypto crates plus `quick-xml` (already a dependency here), and decryption needs no randomness, so the wasm target should be unaffected. Decryption cost is milliseconds.

## Out of scope

- Legacy encrypted `.xls`/`.ppt` (RC4 CryptoAPI): not supported by office-crypto; would keep returning `Encrypted`.
- Encrypted ODF: a different scheme entirely; would keep returning `Encrypted`.

---

*This issue was written with AI assistance (Claude Fable 5).*

Mr-Neutr0n:
Fixes #102

## Summary

Password-protected `.docx`/`.xlsx`/`.pptx` convert when the caller supplies a password, instead of always ending in `Encrypted`. Plaintext files are untouched; without a password nothing changes.

## What changed

- `to_markdown_with_password(path, password)` and `to_markdown_bytes_with_password(bytes, format, password)`: an encrypted OOXML package (OLE container with `EncryptionInfo`/`EncryptedPackage`, detected by the same shape `probe_ole` already classifies) is decrypted via `office-crypto` 0.3 and converted as its plaintext zip, so all existing resource limits apply unchanged. No password / empty password / wrong password → `Encrypted`, exactly as today. Legacy RC4 binary formats (`.doc`/`.ppt`/`.xls`) stay out of scope.
- Wrong-password detection: office-crypto does not check the EncryptionInfo password verifier (a wrong key decrypts to noise with an `Ok` status), so the decrypted payload's zip signature (`PK`) gates success. Documented in `src/package/crypto.rs`.
- CLI: `-p/--password`, with `ANYDOC_PASSWORD` env fallback since argv leaks into shell history and `ps`.
- Node, Python, wasm bindings gain an optional trailing `password` argument (`toMarkdownBytes(bytes, format?, password?)`, `to_markdown_bytes(data, format=None, password=None)`); existing signatures keep working.

## Fixture provenance

`tests/fixtures/encrypted/agile-sha512-docx--errors.docx` is an agile SHA-512 encrypted `.docx` taken from office-crypto's MIT-licensed test suite (password `testPassword`). It carries the corpus sweep's `--errors` annotation for the no-password path; `tests/fixtures/docx/` detection invariants are unaffected.

## Verification

- New integration tests (`tests/encrypted.rs`, 5 cases): convert-with-password returns the expected body; no/wrong/empty password → `Encrypted`; plaintext file with a stray password converts identically.
- Unit tests for the crypto wrapper incl. the non-zip-payload gate.
- Full suite green (288 lib + snapshot corpus + encrypted), clippy clean, fmt clean.
- CLI end-to-end: with `ANYDOC_PASSWORD=testPassword` the fixture converts; without it, `error: document is encrypted`.

This change was prepared with AI assistance under human direction and review.

---
## Summary by cubic
Decrypts password-protected OOXML when a password is supplied and fixes DOCX list numbering to continue across instances that share an abstract, only restarting overridden levels. Previously encrypted OOXML always returned Encrypted; wrong/empty passwords still do, while plaintext files are unchanged.

- New APIs: `to_markdown_with_password(path, password)` and `to_markdown_bytes_with_password(bytes, format, password)` decrypt then convert; exports `is_encrypted_ooxml` and `decrypt_ooxml`. Validates the decrypted payload by opening it as a zip, rejects wrong passwords before applying size limits, enforces a 512 MB decrypted-package budget, and then applies the usual package limits unchanged.
- CLI: `-p/--password` with `ANYDOC_PASSWORD` fallback; a missing `--password` value is a usage error. Password-bearing path conversions route through byte-level conversion after decryption, using content detection with the path extension as the fallback; assets come from the same decrypted bytes.
- Bindings: optional trailing `password` in Node (`toMarkdownBytes(bytes, format?, password?)`), Python (`to_markdown_bytes(data, format=None, password=None)`), and wasm; existing call sites continue to work.
- DOCX numbering: counters key by the resolved abstract so instances sharing one abstract continue a single sequence; only `lvlOverride`/`startOverride` levels restart, and restarts trigger on first use under the overriding instance.

<sup>Written for commit 5e0f81efbf25e3b6df40d963e5dc396982be3135. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/anydoc/pull/130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Mr-Neutr0n:
Rebased/merged onto current `main` and pushed to the same branch (`5e0f81e`). Conflicts are cleared (MERGEABLE). WASM password docs / prior cubic items remain on the branch.

Ready for another look when convenient.

Mr-Neutr0n:
Hi @tomsideguide - a gentle nudge on this OOXML decryption change too; still mergeable and applies cleanly. Happy to update anything, and fine to close if not wanted. Thank you.

PR #148, opened 2026-08-28
fix: bound parser paths reachable from a crafted document

sondt99:
Five resource-exhaustion paths are reachable from untrusted input. One is a hard crash: a ~1.6 MB `.ppt` aborts the process with a stack overflow.

Each is bounded by a constant in `package::limits` now, so it returns `ResourceLimit` like every other attack shape.

### `ppt` — stack overflow (SIGSEGV)

`Extractor::walk` documents itself as *"Iterative container walk over an explicit stack with fixed depth and record-count bounds"* and enforces `MAX_RECORD_DEPTH` against `stack.len()`. But the recovery-mode `NotesContainer` branch called `self.walk(body)` again. The re-entrant call starts a fresh local `stack`, so the depth bound resets to 1 at every level and never fires — nesting ends up capped only by `MAX_RECORDS` (16M), orders of magnitude past what the real stack survives.

`recovering` is attacker-controlled: any unusable persist directory sets it.

Reproduced on `261fc25`:

```
test formats::ppt::tests::nested_notes_in_recovery_hit_the_depth_bound ... FAILED
thread '...::deeply_nested_notes_do_not_overflow_the_stack' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

The descent is a stack push carrying an end-of-segment flag now, so the `end_segment` / `current_is_notes` ordering is byte-for-byte what the recursive version did.

### `rtf` — stack overflow

Math zones build their `Element` tree through `MathState::open_group`, not `parse_xml`, so `MAX_XML_DEPTH` never applied to it. `close_groups` nests each group into its parent, making tree depth equal brace depth, and the OMML serializer (`walk_children`/`walk_elem`) recurses once per level — as does dropping the tree. Capped at construction, since a guard in the serializer would not save the recursive `Drop`.

Separately the group stack retained a `CharState` per `{` with no cap: ~10 MB of `{` → ~1 GB RSS. Now `MAX_RTF_GROUP_DEPTH`.

### `doc` — OOM and allocator abort

FKP page numbers in the PLC are masked to 22 bits and nothing stops every entry pointing at the same page, so the entry count did not bound the run list — a few MB of table stream expanded into tens of GB. Bounded by `MAX_DOC_RUNS`; formatting degrades past it, text extraction is unaffected.

`parse_plf_lfo` also sized a `Vec` straight from a raw `u32` header field, so `0xFFFFFFFF` asked the allocator for ~68 GB and hit `handle_alloc_error` → `abort()`. Clamped to what the buffer can hold; the existing loop already bounded the real count.

### `sheet` — OOM

`formatCode` has no length bound anywhere and `parse_section` materializes several vectors per character, so a compressed archive entry amplified into GBs without a worksheet cell being involved (`Styles::read` parses every `cellXfs/xf` unconditionally). Capped by `MAX_NUMBER_FORMAT_BYTES`, falling back to General.

### Also

`#![forbid(unsafe_code)]` at the crate root. The crate already had zero `unsafe` — this makes the property enforced rather than incidental. (Scoped to this crate; the parsing dependencies are unaffected.)

### Tests

Four regression tests, two of which fail on the parent commit — the `.ppt` one by aborting the test binary.

`cargo fmt --check` clean · `cargo clippy --all-targets -- -D warnings` exit 0 · 300 tests pass · workspace incl. Python and WASM bindings compiles.

Happy to split this into per-format PRs, or to move the crash to a private advisory instead if you'd rather it not sit in a public diff — there is no `SECURITY.md`, and since this is a DoS in safe Rust rather than memory corruption I defaulted to the normal PR flow.

---
## Summary by cubic
Fixes five resource-exhaustion paths reachable from crafted documents, including a `~1.6 MB` `.ppt` that aborts the process with a stack overflow. Each path now returns `ResourceLimit` instead of crashing or exhausting memory, matching how other attack shapes are handled.

**Bug Fixes**
- The recovery-mode notes container in `ppt` now descends via the shared bounded stack instead of re-entering `walk`, which reset the depth bound to 1 at every level and left nesting capped only by the 16M record limit.
- RTF math-zone nesting is capped at construction, since the recursive `Drop` would defeat a serializer-side guard; the group stack also gets its own cap to bound retained character-state snapshots.
- `.doc` FKP formatting runs are bounded, and `parse_plf_lfo` clamps a raw header count to the buffer length instead of reserving tens of gigabytes from a value like `0xFFFFFFFF`.
- `formatCode` longer than 4096 bytes falls back to General rather than amplifying a compressed archive entry into gigabytes.
- `#![forbid(unsafe_code)]` makes the crate's existing safe-only property enforced rather than incidental.

<sup>Written for commit f8c87cc73ab6651b8e1a9355e8332902c93e182a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/anydoc/pull/148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/formats/doc/lists.rs
+++ b/src/formats/doc/lists.rs
@@ -194,7 +194,12 @@ fn parse_plf_lfo(plf: &[u8], by_lsid: &HashMap<u32, [LevelDef; LEVELS]>, lists: 
     let Some(count) = get_u32(plf, 0).map(|v| v as usize) else {
         return;
     };
-    let mut lfos: Vec<(u32, usize)> = Vec::with_capacity(count);
+    // `count` is a raw header field: reserve only what the buffer could
+    // actually hold, or a lie like 0xFFFFFFFF asks the allocator for tens of
+    // gigabytes and aborts the process. The loop below still bounds the real
+    // count, since it stops as soon as a record runs past the end.
+    let capacity = count.min(plf.len().saturating_sub(4) / LFO_SIZE);
+    let mut lfos: Vec<(u32, usize)> = Vec::with_capacity(capacity);
     let mut pos = 4;
     for _ in 0..count {
         let Some(record) = plf.get(pos..).and_then(|rest| rest.get(..LFO_SIZE)) else {
--- a/src/formats/doc/mod.rs
+++ b/src/formats/doc/mod.rs
@@ -447,6 +447,13 @@ fn parse_fkps(
     }
     let n = (plc.len() - 4) / 8;
     for i in 0..n {
+        // Nothing stops every entry pointing at the same page, so `n` does
+        // not bound the run count: without this a few megabytes of table
+        // stream expand into tens of gigabytes of `Run`.
+        if runs.len() >= limits::MAX_DOC_RUNS {
+            log::debug!("doc: formatting-run budget reached; later runs are ignored");
+            break;
+        }
         let Some(pn_raw) = get_u32(plc, (n + 1) * 4 + i * 4) else {
             continue;
         };
--- a/src/formats/ppt/mod.rs
+++ b/src/formats/ppt/mod.rs
@@ -376,11 +376,22 @@ impl Extractor {
     /// Iterative container walk over an explicit stack with fixed depth and
     /// record-count bounds — nesting or record counts beyond any real
     /// presentation are attack shapes and hard-fail.
+    ///
+    /// Every descent must push onto `stack`. Nothing in here may call `walk`
+    /// again: the depth bound reads `stack.len()`, so a re-entrant call would
+    /// start a fresh stack, reset the bound to 1 and leave nesting unbounded
+    /// all the way to a stack overflow.
     fn walk(&mut self, data: &[u8]) -> Result<(), ConvertError> {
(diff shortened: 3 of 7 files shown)


GeorgMelinius:
I addressed all three unresolved Cubic findings in https://github.com/sondt99/anydoc/pull/2 (commit cddc9cbc3fb760b4659f7ae3d605ef79a093a66b):

- clarified that selected limits intentionally degrade gracefully
- rewrote the 200,000-level PPT fixture builder from quadratic copying to one linear allocation
- removed the RTF ignored-destination marker so the math test enters MathState::open_group

Verified with Rust 1.88 using cargo fmt --all --check, cargo clippy --workspace --all-targets --all-features -- -D warnings, and cargo test --locked. All 290 unit tests, the robustness test, and 9 snapshot tests passed (1 local-samples test ignored).

Merging the companion PR will update this PR branch with the fixes.

Issue #63, opened 2026-08-07
Expose embedded image assets in markdown output


adiberk:
## Context

When processing DOCX files, `to_document()` correctly extracts embedded images into `doc.assets` (with raw bytes, media type, and origin part). However, these images don't appear in the markdown output from `to_markdown_bytes()` — there are no `![alt](...)` references, and the inline block model doesn't surface image inlines.

## Current behavior

```python
doc = anydoc.to_document(docx_bytes)
print(len(doc.assets))  # 1 (608KB PNG found)

md = anydoc.to_markdown_bytes(docx_bytes)
print("![" in md)  # False (no image references)
```

The asset data is extracted and available, but the markdown doesn't reference it.

## Request

It would be useful if `to_markdown_bytes()` emitted image references (e.g., `![alt text](asset:0)` or `![](embedded:image0.png)`) at the positions where images appear in the document. The actual bytes are already in `doc.assets` — consumers could then decide whether to inline them as data URIs, upload them, or send them to a vision model for descriptions.

This would also apply to XLSX and PPTX files that contain embedded images/charts.

## Workaround

Using `to_document()` to get the asset bytes and processing them separately works, though without positional information it's hard to associate an image with the surrounding content.

mosabalrsaheed:
@tomsideguide any ETAs for the PR to be merged for this?! 👍

justinzyw:
is it possible to at least put the asset_id to where it is in the markdown so that later I can somehow process the asset to put the content of the asset back to where it should be?

PR #151, opened 2026-08-31
fix(xlsx): retain embedded worksheet images

### LINKED ISSUE #139: XLSX embedded images are missing from Document.assets
## Summary

`anydoc.to_document(..., "xlsx")` extracts worksheet cell content but omits embedded worksheet images from `Document.assets`.

This differs from #63: that issue reports assets being present for DOCX but absent from rendered Markdown. In this case, the XLSX asset is already missing from the shared `Document` model, so consumers cannot preserve the image bytes or create their own Markdown reference.

## Environment

- `firecrawl-anydoc`: 0.2.3
- Python: 3.11.15
- OS: Windows

## Minimal reproduction

```python
import tempfile
from pathlib import Path

import anydoc
from openpyxl import Workbook
from openpyxl.drawing.image import Image

PNG = bytes.fromhex(
    "89504e470d0a1a0a0000000d49484452000000010000000108060000001f15c489"
    "0000000d49444154789c6360f8cfc000000301010018dd8db10000000049454e44"
    "ae426082"
)

with tempfile.TemporaryDirectory() as directory:
    root = Path(directory)
    image_path = root / "evidence.png"
    workbook_path = root / "with-image.xlsx"
    image_path.write_bytes(PNG)

    workbook = Workbook()
    worksheet = workbook.active
    worksheet["A1"] = "visual evidence"
    worksheet.add_image(Image(str(image_path)), "B2")
    workbook.save(workbook_path)
    workbook.close()

    payload = workbook_path.read_bytes()
    document = anydoc.to_document(payload, "xlsx")

    print("format:", anydoc.format_from_path(workbook_path))
    print("assets:", len(document.assets))
    print("blocks:", len(document.blocks))
    print(anydoc.to_markdown_bytes(payload, "xlsx"))
```

## Actual result

```text
format: xlsx
assets: 0
blocks: 1
|  |
| --- |
| visual evidence |
```

The cell content is extracted, but the embedded PNG is absent from `document.assets`.

## Expected result

`document.assets` should contain the embedded worksheet image, including its bytes and media type, consistently with the documented shared document model contract for embedded images.

Positional image rendering in Markdown can remain tracked separately in #63; exposing the image in `Document.assets` is sufficient for consumers to persist it themselves.

Fanfulla:
This fixes the XLSX side of #139.

The parser now follows the worksheet -> drawing -> image relationship chain and stores embedded image bytes in `Document.assets`, using the existing `AssetSink` behavior for byte limits and deduplication. Each asset keeps its media type and original package path.

I kept Markdown output out of scope. This change only makes the shared document model complete for XLSX files. Positional Markdown references are still tracked in #63.

Tests:

- added a regression test with a minimal XLSX containing an embedded image
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test --locked`

Fixes #139

### DIFF
--- a/src/formats/docx/content.rs
+++ b/src/formats/docx/content.rs
@@ -66,7 +66,7 @@ impl<'a, 'b> Ctx<'a, 'b> {
     /// part. Failures degrade (log + `None`) per the unified policy;
     /// resource-limit errors always propagate.
     fn rel_part(&self, rel_id: &str) -> Result<Option<RelTarget>, ConvertError> {
-        rel_target_bytes(self.pkg, &self.rels, &self.base_part, rel_id)
+        rel_target_bytes(&mut self.pkg.borrow_mut(), &self.rels, &self.base_part, rel_id)
     }
 
     fn add_asset(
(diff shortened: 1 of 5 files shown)


Fanfulla:
Fixed in 11c8c14. `rel_target_bytes` and `rel_image_source` now accept mutable references directly, so XLSX, DOCX, and PPTX use the same relationship resolution and asset retention path.

I reran `cargo fmt --all --check`, Clippy with warnings denied, and the locked test suite.

PR #53, opened 2026-08-07
feat: support standalone HTML documents

xianjianlf2:
## Summary

Closes #52.

- Add standalone HTML detection from `<!doctype html>` or an `<html>` root, including HTML exports saved with a `.doc` extension.
- Reuse the EPUB HTML-to-model walker for headings, formatting, links, tables, lists, CSS visibility, and external images.
- Treat HTML void elements such as unclosed `<meta>` and `` as empty elements in the shared DOM parser.
- Expose `html` / `.html` / `.htm` through the Rust, Node, Python, and WASM APIs and update docs/tests.
- MHTML remains out of scope for this change.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test --locked`
- `npm test`
- `python -m unittest discover -s python/tests`
- `node --test wasm/test.mjs`

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/anydoc/pr/53"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788679080&installation_model_id=11126&pr_number=53&repository=firecrawl%2Fanydoc&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Fanydoc%2Fpull%2F53&signature=bc88c6c8b7a09cd9d148b72d1c758a488f590cf5e2e512cb58914db923d94e39"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

Issue #52, opened 2026-08-07
Feature request: Support for HTML / MHTML files (e.g. Jira exports)


thvroyal:
## Problem

Jira exports issues as standalone HTML files wrapped with `Content-Type: application/vnd.ms-word`, often saved with a `.doc` extension. These files are plain HTML documents (not OLE2 compound binaries), and anydoc currently rejects them as malformed:

```
malformed document: not an OLE2 compound file: Invalid CFB file (wrong magic number): [3c, 21, 44, 4f, 43, 54, 59, 50]
```

The magic bytes `3c 21 44 4f 43 54 59 50` = `<!DOCTYPE` -- it is a valid HTML file.

## Use case

Many enterprise tools (Jira, Confluence, SharePoint) export content as HTML or MHTML (`.mht`, `.mhtml`) files. Being able to convert these to Markdown via anydoc would be valuable for document pipelines that ingest exports from these tools.

## Proposed solution

Add support for standalone HTML files as an input format. The EPUB parser already handles HTML internally, so much of the parsing logic may be reusable.

At minimum, recognizing the `<!DOCTYPE html>` header (or `<html>` root element) and routing through the existing HTML-to-model pipeline would cover the Jira export case. Full MHTML (multipart MIME with embedded resources like images) support would be a bonus.

## Example file

[jira_export_ticket.doc](https://github.com/user-attachments/files/30816559/AI1SOLHPCHAT-98.doc)

marcellmanfrin:
I did a quick pass over the current architecture and this looks substantially easier than implementing HTML support from scratch, because `src/shared/html.rs` already contains the semantic HTML/XHTML → `Document` conversion used by EPUB.

A practical implementation path could be:

## 1. Add standalone HTML as a first-class format

The main missing piece is a tolerant HTML5 parser. EPUB currently receives well-formed XHTML and can rely on the XML tree, but real-world `.html` files are not necessarily XML-compatible (`<br>`, `<img>`, omitted closing tags, malformed tables, etc.).

`html5ever` looks like a strong fit here: mature, browser-grade HTML5 parsing in Rust, without requiring Chromium/Playwright or JavaScript execution.

Proposed pipeline:

```text
HTML bytes
  -> charset detection/decoding
  -> html5ever
  -> DOM/tree
  -> existing shared::html::to_blocks(...)
  -> Document
  -> Markdown
```

This keeps the feature aligned with anydoc's current model: deterministic document parsing rather than browser rendering.

[...]

PR #164, opened 2026-09-04
feat(eml): read RFC 5322 email messages

itsaaronngan:
Closes half of #128: `.eml` only. `.msg` is #160, from @vaibhavdabas16 — independent frontends, either order.

## Shape

Hand-rolled MIME on `std` + `encoding_rs`, no new dependencies. Follows `csv.rs`: `parse(bytes) -> Result<Document, ConvertError>`, produces the document model, does not touch the Markdown writer.

- RFC 5322 headers, continuation lines unfolded
- multipart traversal, depth-capped, prefers `text/plain`
- quoted-printable and base64 transfer-decoding
- RFC 2047 encoded-words, both B and Q
- charset via `encoding_rs`, mirroring `csv.rs::decode`
- Subject as H1, From/To/Cc/Date as a metadata paragraph
- extension-only detection, no `detect.rs` entry, per the CSV precedent for signature-less text formats

Registered in the node, wasm and python enums and in `node/cli.js` FORMATS.

A body that is `text/html` with no `text/plain` alternative returns `Unsupported` rather than emitting markup as prose. That case wants #52 or #53 landing first; I did not want to guess at HTML here.

## Validation

142 real `.eml` files, Outlook/Exchange via Mimecast: 141 convert, median 4.0 ms.

The one failure is a calendar invite whose `text/plain` part is 2 bytes and whose content is `text/calendar`. It returns `Unsupported` rather than emitting the 2 bytes. Correct as far as I can tell, but it is a judgement call — happy to extract it instead if you would rather.

Running the whole corpus, rather than spot-checking, is what made this worth trusting. It surfaced an RFC 2047 bug I would not have invented a test for: `decode_one_word` looked for the terminator `?=` from the start of the word, but a payload's own `=XX` escapes can put a literal `?=` inside it, so the search landed early and parsing bailed, leaving raw encoded words in the subject. Fixed by locating the terminator after both `?` separators, with a regression test. Post-fix the corpus has zero undecoded escapes and zero U+FFFD.

That corpus is private, so the fixtures are synthetic: seven handmade files under `tests/fixtures/eml/`, plus the `fixtures_detect_from_content` entry. Writing them turned up four more bugs I had missed against real mail, the worst being that a whitespace-only line (one space, not empty) did not break a paragraph, so it rendered as a hard break with nothing after it. 3267 occurrences across the corpus, now zero.

Fuzzed as well, since the parser walks attacker-shaped bytes: 3,088,842 executions over the two seeds, no crashes, no hangs, no leaks.

Checks, on this branch against v0.2.4:

- `cargo test --locked` — 309 passed, 1 ignored (13 new unit tests)
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo clippy -p anydoc-wasm --target wasm32-unknown-unknown -- -D warnings`
- `cargo fmt --all --check`
- `cargo +nightly fuzz run eml` — 3,088,842 executions, clean

## Two notes for review

**`detect.rs`.** I left `.eml` on extension-only detection, reading the CSV precedent in the README ("CSV has no such marker, so the extension or an explicit format names it instead") as the rule for signature-less text formats. #160 chose a content-sniff entry for `.msg`, which is a different case — OLE has a real signature. If you would rather `.eml` sniffed too, say so and I will add it; I did not want to invent a heuristic that `detect.rs` warns against.

**Shared helper.** If both this and #160 land, each frontend carries its own envelope-and-body rendering. @vaibhavdabas16 offered to factor that out as a follow-up once the second one is in, and that is the right shape — he has the clearer view of both. I am happy to take it instead if he would rather not.

---
## Summary by cubic
Adds RFC 5322/MIME `.eml` conversion, closing the `.eml` half of #128. Previously unsupported, messages now convert to Markdown with the Subject as an H1, sender metadata, and a `text/plain` body preferred over `text/html`.

**Behavior**
- Supports folded headers, RFC 2047 encoded words, quoted-printable and base64 bodies, charsets, and nested multipart messages without new dependencies.
- Excludes attachments, recovers missing boundaries as flat text, and returns `Unsupported` for HTML-only messages.
- Uses extension or explicit-format selection because EML has no content signature, and exposes the format in the Rust, Node, Python, WASM, and CLI APIs.
- Handles padded closing boundaries without consuming the epilogue and preserves whitespace before malformed encoded-word candidates.

**Validation**
- Adds fixtures, snapshots, unit tests, and fuzz coverage; 141 of 142 real Outlook/Exchange messages convert with a 4 ms median, and 3M+ fuzz executions completed cleanly.

<sup>Written for commit af00979796307ee5059b003927a3fde1f4139d00. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/anydoc/pull/164?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![License: MIT](https://img.shields.io/badge/license-MIT-blue.svg)](LICENSE)
 [![skills.sh](https://skills.sh/b/firecrawl/anydoc)](https://skills.sh/firecrawl/anydoc)
 
-Fast Rust library that converts documents (Word, PowerPoint, Excel, OpenDocument, RTF, EPUB, CSV, and PDF) into clean GitHub-Flavored Markdown. Includes bindings for [Node.js](node/README.md), [Python](python/README.md), and the [browser](wasm/README.md) (WebAssembly).
+Fast Rust library that converts documents (Word, PowerPoint, Excel, OpenDocument, RTF, EPUB, CSV, email, and PDF) into clean GitHub-Flavored Markdown. Includes bindings for [Node.js](node/README.md), [Python](python/README.md), and the [browser](wasm/README.md) (WebAssembly).
 
 Built by [Firecrawl](https://firecrawl.dev) to turn any office document into LLM-ready Markdown in single-digit milliseconds, with one consistent output no matter which format goes in. It powers [Firecrawl Parse](https://firecrawl.dev/parse), so if you'd rather not run it yourself, the hosted API gives you the same conversion plus our OCR models for the scanned pages anydoc can't read on its own.
 
@@ -167,6 +167,7 @@ Only documents that need OCR leave the machine, and the whole document goes, sin
 | Rich Text Format | `.rtf`                                                     |
 | EPUB             | `.epub`                                                    |
 | CSV              | `.csv`                                                     |
+| Email            | `.eml`                                                     |
 | PDF              | `.pdf`                                                     |
 
 ## Benchmark
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ Speed is one warm conversion per document on a Ryzen 9 9950X3D (Windows 11, 64 G
 
 ## Format detection
 
-The format is read from the file content, using the marker its specification designates: the PDF header, the RTF open group, OLE stream names, the ZIP package mimetype and content types. CSV has no such marker, so the extension or an explicit format names it instead.
+The format is read from the file content, using the marker its specification designates: the PDF header, the RTF open group, OLE stream names, the ZIP package mimetype and content types. CSV and EML have no such marker, so the extension or an explicit format names them instead.
 
 ```rust
 Format::from_bytes(&bytes); // Some(Format::Docx), or None when nothing matches
@@ -262,7 +262,7 @@ document bytes
   ├─► format detection      → content markers, not the extension
   │
   ├─► format parser          → one per format (doc, docx, ppt, pptx, xls,
-  │                            xlsx, odt/ods/odp, rtf, epub, csv)
+  │                            xlsx, odt/ods/odp, rtf, epub, csv, eml)
   │         │
   │         └─► Document     → shared model: blocks, inlines, tables,
   │                            footnotes, assets
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -84,3 +84,9 @@ name = "numfmt"
 path = "fuzz_targets/numfmt.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "eml"
+path = "fuzz_targets/eml.rs"
+test = false
+doc = false
--- a/fuzz/README.md
+++ b/fuzz/README.md
@@ -5,6 +5,10 @@
 `fuzz/seeds/` holds checked-in starting inputs; `fuzz/corpus/` is the working
 directory libfuzzer writes to and is not checked in.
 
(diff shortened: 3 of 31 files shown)


itsaaronngan:
Thanks — six of the seven are real and are fixed in c47d71d, each with a regression test that I checked fails against the previous commit rather than passing vacuously.

The two that mattered were genuine correctness bugs, and both are cases my 142-message corpus simply never contained:

- **Attachment displacing the body (P1).** Right, and worse than described: because the `text/plain` pass runs before the nested-multipart pass, a `multipart/mixed` of `[multipart/alternative, text/plain attachment]` returned the attachment and never reached the real message. Now excluded via RFC 2183 disposition.
- **Boundary prefix (P2).** Also right, and this one lost content silently — a body line beginning `--abcExtra` was read as a delimiter and everything after it was dropped with no error. Now matched per RFC 2046 5.1.1: the boundary exactly, optionally closed with `--`, then only whitespace.
- **Adjacent encoded-words (P2).** Correct per RFC 2047 6.2. Whitespace between a word and ordinary text is still preserved.

The three P3 doc points are fixed too — EML is now named alongside CSV in the `Format`/`from_bytes`/`to_markdown_bytes` doc comments, the CLI's stdin help, and the README's detection paragraph and parser diagram.

**One I do not think is valid: the base64 padding finding (P2, `eml.rs:344`).** The claim is that misplaced or incomplete padding is accepted and "can produce corrupted body text". Decoding `aGVs=bG8gd29ybGQ=` as a quoted body gives `hello world`, not corruption — skipping `=` wherever it appears is lenient, but it is lenient in the direction Postel's law wants for mail, and I could not construct an input where it produces wrong text rather than simply tolerating a malformed one. Returning `None` there would send valid-enough real messages down the recovery path instead. Happy to be shown a counter-example if one exists.

Re-fuzzed after the changes: 1,898,113 executions, no crashes, no hangs, no leaks. Full suite is 312 passing, clippy clean on the workspace and the wasm target, fmt clean.

itsaaronngan:
Both valid, both mine, both fixed in af00979.

- **Padded closing delimiter.** Right — I widened `is_delimiter` to accept the trailing whitespace RFC 2046 5.1.1 permits but left the close check testing the raw line, so `--B--  ` opened a part instead of closing the multipart and the epilogue was parsed. `is_delimiter` now reports which form it matched, so there is one decision rather than two that can disagree.
- **Whitespace before a malformed token.** Also right. The gap is now resolved after the candidate decodes, so it is only suppressed between two genuinely decoded words.

Worth noting for anyone reading the history: my first regression test for the padded-close case passed against the unfixed code. The fixture put the `text/plain` part before the close, so the correct body was found regardless and the test never reached the bug. I only caught that by reverting the fix and checking the test actually went red. Both new tests are verified that way now, as are the three from the previous round.

Re-fuzzed after these changes: 1,652,818 executions, no crashes, no hangs, no leaks. Suite is 314 passing, clippy clean on the workspace and the wasm target, fmt clean.

The base64 padding item is still open from my side — I could not construct an input where lenient `=` handling yields wrong text rather than merely tolerating a malformed payload, and returning `None` there would push valid-enough real messages down the recovery path. Genuinely happy to be shown a counter-example.

